PR #4, opened 2017-08-02
Add in gradle source and build config

loosebazooka:
Usage : This CL adds in three basic tasks to the default tasks `minikubeStart`, `minikubeStop` and `minikubeDelete`.

The minikube plugin has an extension with a single configuration parameter (minikube) for users to configure the location of minikube for use with all minikube tasks. 
```groovy
minikube {
  minikube = "/path/to/minikube" // default is "minikube"
}
```
Task specific flags are configured on the tasks themselves, all minikube tasks are of the type `MinikubeTask` and all share the same kind of configuration.
- `minikube` (String) : path to minikube executable which should be set by the `minikube` extension.
- `command` (String) : start/stop/whatever (users probably shouldn't be editing this for default commands)
- `flags` (String[]) : any minikube flags, **this is the only one users should edit**

```groovy
minikubeStart {
  flags = ["--vm-driver=none"]
}
```

This CL also allows users to add in any custom minikube task, which will be configured with the minikube extension value for `minikube` if set.

```groovy
task minikubeBonk(type: com.google.cloud.tools.minikube.MinikubeTask) {
  command = "bonk"
  flags = ["--some-flag"]
}
```

The plan is to eventually add an extension that allows users to share `minikube` and some common `flags` if they so chose to. But this CL doesn't cover that.


loosebazooka:
@r2d4, @aaron-prindle fyi

minikube-gradle-plugin/build.gradle
```diff
@@ -1,5 +1,133 @@
...
+}
+
+
+/* GOOGLE JAVA FORMAT */
+check.dependsOn verifyGoogleJavaFormat
+/* GOOGLE JAVA FORMAT */
+
+
+/* CHECKSTYLE */
+checkstyle {
```
patflynn: can you exclude the java files from this?  google-java-format should handle that part.

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/gradle/MinikubeTask.java
```diff
@@ -0,0 +1,114 @@
...
+  @Input
+  public String[] getFlags() {
+    return flags;
+  }
+
+  public void setFlags(String[] flags) {
+    this.flags = flags;
+  }
+
+  /** Task entry point. */
```
patflynn: Is this passing lint?

minikube-gradle-plugin/src/test/java/com/google/cloud/tools/minikube/gradle/MinikubePluginTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/*
+ * This Java source file was auto generated by running 'gradle init --type java-library'
+ * by 'appu' at '6/8/17 3:08 PM' with Gradle 2.0
+ *
+ * @author appu, @date 6/8/17 3:08 PM
```
patflynn: update javadoc

minikube-gradle-plugin/build.gradle
```diff
@@ -1,5 +1,133 @@
...
+    doLast {
+      pom {
+        project {
+          name 'App Engine Gradle Plugin'
+          description 'This Gradle plugin provides tasks to build and deploy Google App Engine applications.'
+          url 'https://github.com/GoogleCloudPlatform/app-gradle-plugin'
+          inceptionYear '2016'
+
+          scm {
+            url 'https://github.com/GoogleCloudPlatform/app-gradle-plugin'
```
r2d4: Should these be updated to `https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java`?

minikube-gradle-plugin/src/test/java/com/google/cloud/tools/minikube/gradle/MinikubePluginTest.java
loosebazooka: done

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/gradle/MinikubeTask.java
loosebazooka: It was checkstyle that was failing here. Removed it.

minikube-gradle-plugin/build.gradle
loosebazooka: done

loosebazooka:
@r2d4 @patflynn ptal?

patflynn:
@loosebazooka it'd be cool to have an issue or a commit message that describes the UX that this PR enables.

loosebazooka:
Right, that's a good point. Will add that in

loosebazooka:
updated description.

minikube-gradle-plugin/src/main/resources/META-INF/gradle-plugins/com.google.cloud.tools.minikube.properties
```diff
@@ -0,0 +1 @@
+implementation-class=com.google.cloud.tools.minikube.gradle.MinikubePlugin
```
r2d4: nit: newline

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/gradle/MinikubeTask.java
```diff
@@ -0,0 +1,111 @@
...
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.TaskAction;
+
+/** Generic Minikube task. */
+public class MinikubeTask extends DefaultTask {
+
+  /**
+   * The minikube executable, can include the full path to minikube, default assumes it is on path.
+   * I'm not sure this works on windows, ask someone about this.
+   */
```
r2d4: I think we install as `minikube.exe` on windows.  Not sure if this complicates things a bit on your end. @aaron-prindle knows more.

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/gradle/MinikubePlugin.java
```diff
@@ -0,0 +1,27 @@
...
+  private void createStartTask() {
+    MinikubeTask task = project.getTasks().create("minikubeStart", MinikubeTask.class);
+    task.setCommand("start");
+  }
+
+  private void createStopTask() {
+    MinikubeTask task = project.getTasks().create("minikubeStop", MinikubeTask.class);
+    task.setCommand("stop");
+  }
+}
```
r2d4: if its not too much work, the `delete` is the other super basic/common command.  that would complete the triforce.

loosebazooka: will do.

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/gradle/MinikubeTask.java
loosebazooka: @aaron-pindle @r2d4 any idea here?

I think we might need to launch using "cmd \c minikube.exe"

minikube-gradle-plugin/src/main/resources/META-INF/gradle-plugins/com.google.cloud.tools.minikube.properties
loosebazooka: done

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/gradle/MinikubeTask.java
r2d4: I think it needs to be launched in powershell?  I'm actually not too sure on this one.  Feel free to leave it out for now if its tricky.

loosebazooka:
@r2d4 PTAL

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/gradle/MinikubeTask.java
aaron-prindle: The binary is `minikube.exe` on Windows.  If users try to run minikube using HyperV then it does need to be run through Powershell but for Virtualbox, any shell will work.

r2d4: I think we should just default to powershell on windows then, so that it supports both hypervisors.

loosebazooka:
@r2d4 added in some thing to make configuration better, see PR description.

PR #8, opened 2017-09-28
Release v1.0.0 alpha.1

loosebazooka:
Prepare for alpha.1 launch.


coollog:
So, we are releasing a version called v1.0.0-alpha.1? And then the next version would be v1.0.0?

Issue #7, opened 2017-09-27
minikubeDockerBuild


loosebazooka:
run a docker build into the minikube env.

command line equivalent 
```
> eval $(minikube docker-env)
> docker build ....
```

Probably exposed to the user as `minikubeDockerBuild`, unfortunately there's no official docker plugin to link to, so we might have to use the docker java api directly to initiate a build.

loosebazooka:
you can use https://github.com/kubernetes/minikube/pull/1792 to obtain the docker config for minikube

coollog:
Starting with getting the minikube docker env vars, will take these steps:

1) Create a shared utility class called CommandExecutor that can run a shell command and get the output.
2) Use this CommandExecutor in a new DockerBuildTask to retrieve the minikube docker env vars.

coollog:
Then will take these steps:

PR #10, opened 2017-09-29
Moves command execution to utility class CommandExecutor.

coollog:
Step 1 of #7 .


minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/util/CommandExecutor.java
```diff
@@ -0,0 +1,163 @@
...
+  public CommandExecutor() {}
+
+  public CommandExecutor setLogger(Logger logger) {
+    this.logger = logger;
+    return this;
+  }
+
+  @VisibleForTesting static final int TIMEOUT_SECONDS = 5;
+
+  @VisibleForTesting
```
loosebazooka: Don't use this annotation from `org.gradle.internal.impldep.com.google.common.annotations.VisibleForTesting`, we don't pull in the dependency ourselves to guava, so just either leave it in but commented, or take it out.

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/util/CommandExecutor.java
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Helper function that runs a process on an {@code ExecutorService} and stores the output as a
+   * list of lines
+   *
+   * @param process the process to run
+   * @param executor the {@code ExecutorService} that executes the process
+   * @param output the {@code List} to store the output in
+   * @return the exit code of the process
+   */
+  private int runProcessWithExecutor(Process process, ExecutorService executor, List<String> output)
```
loosebazooka: I don't think you need to separate this function out, it seems like it can just sit in `run` with the rest of the stuff.

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/util/CommandExecutor.java
```diff
@@ -0,0 +1,163 @@
...
+      throws InterruptedException {
+    executor.execute(outputConsumerRunnable(process, output));
+
+    return process.waitFor();
+  }
+
+  private void shutdownExecutor(ExecutorService executor) {
```
loosebazooka: Same thing here about separation, perhaps leave this in `run`?

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/util/CommandExecutor.java
```diff
@@ -0,0 +1,163 @@
...
+
+    // Stops the build if the command fails to do something, we may want to make this configurable.
+    if (exitCode != 0) {
+      throw new GradleException("command exited with non-zero exit code : " + exitCode);
+    }
+
+    return output;
+  }
+
+  private Process buildProcess(List<String> command) throws IOException {
```
loosebazooka: Can probably just leave in `run`. We're not testing this anyway, it feels slightly harder to follow.

minikube-gradle-plugin/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
loosebazooka: Nit : looks like there's no new line at the end of this file?

coollog: Okay added the newline.

PR #12, opened 2017-10-03
Allows for environment variables to be used for CommandExecutor.

coollog:
Step 2 of #7 .


loosebazooka:
Seems a little complicated. Is it only like this because you want to parse the output of minikube docker-env?

coollog:
Yes, this is so I can parse the output of https://github.com/kubernetes/minikube/pull/1792 from the "KEY=VALUE" form into a `Map<String, String>` I can directly just `putAll` into the environment of the `ProcessBuilder`.

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/util/EnvironmentVariableMap.java
```diff
@@ -0,0 +1,86 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.minikube.util;
+
+import java.util.HashMap;
+import java.util.List;
+
+/** Represents a environment map from environment variable keys to values. */
+public class EnvironmentVariableMap extends HashMap<String, String> {
```
loosebazooka: This class itself might be a little overkill for what we're trying to do. 
- Perhaps a simple parser that reads a list of strings and outputs a map is all we need. MinikubeEnvParser or something. Then we get a standard Map instance to manipulate in other places.
- Not super on board with the idea of extending Map. Making this an input parameter to "setEnvironment" doesn't seem to allow us to use a simple Map in CommandExecutor if we wanted to. The extended class itself doesn't offer too much in terms of functionality and would probably (subjective of course) more readable if it just kept an instance of a HashMap.
- I would probably avoid the custom runtime exceptions, if they're not going to be checked then we might as well use one of the well known existing Runtime exceptions like `IllegalArgumentException`.

coollog: Okay changed the class to MinikubeDockerEnvParser to just be a static parser.

PR #13, opened 2017-10-03
Adds DockerBuildTask.

coollog:
Step 3 and 4 of #7 .


minikube-gradle-plugin/README.md
```diff
@@ -10,16 +10,23 @@ It exposes the following tasks
 - `minikubeStart`
 - `minikubeStop`
 - `minikubeDelete`
+- `minikubeDockerBuild` - runs `docker build` in the minikube Docker environment
 
-It exposes the `minikube` configuration extension.
+It exposes the `minikube` and `docker` configuration extension.
```
loosebazooka: So it might be a little risky to expose a `docker` extension. It might clash with other docker plugins with an error like "Cannot create 'docker' extension, because it already exists"

loosebazooka: I you want, you can create the `docker` extension under the `minikube` extension instead of on the project.

loosebazooka: or just add `docker` it under the minikube extension as a configuration parameter. <-- this might be what I would do.

coollog: Should we just have the docker client configuration be in minikube extension as well then?

loosebazooka: yeah, so it looks like ?
```
minikube {
  minikube 'some/path'
  docker = 'some/path'
}

Comment (coollog) [reply minikube-gradle-plugin/README.md]: Yea, I'll change it to be that way

File: minikube-gradle-plugin/README.md [hunk 2]
```diff
@@ -10,16 +10,23 @@ It exposes the following tasks
...
+
+docker {
+  docker = // path to Docker, default is "docker"
+}
 ```
 
-Task specific flags are configured on the tasks themselves, all minikube tasks are of the type `MinikubeTask` and all share the same kind of configuration.
+Task specific flags are configured on the tasks themselves.
+ 
+All `minikube` tasks are of the type `MinikubeTask` and all share the same kind of configuration.
```
loosebazooka: all tasks except minikubeDockerBuild?

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/DockerBuildTask.java
```diff
@@ -0,0 +1,141 @@
...
+    List<String> minikubeDockerEnvCommand =
+        Arrays.asList(minikube.get(), "docker-env --shell none");
+    List<String> dockerEnv =
+        commandExecutorFactory.createCommandExecutor().run(minikubeDockerEnvCommand);
+
+    Map<String, String> environment;
+    try {
+      environment = MinikubeDockerEnvParser.parse(dockerEnv);
+    } catch (IllegalArgumentException ex) {
+      throw new GradleException("Exception while processing minikube's Docker environment");
```
loosebazooka: can you add ex to the context?
new GradleException(message, ex)?

loosebazooka: you might be able to just let this runtime exception fall through and gradle auto converts it.

loosebazooka: I just looked at the message that the exception in minikubeDockerEnv parser throws, it might not be as helpful as what you've done here. Maybe we can update that message to be more informative rather than the generic messaging it currently outputs (since the class already names itself as though it is parsing minikube env)

https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/blob/master/minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/util/MinikubeDockerEnvParser.java#L42

coollog: Okay, made the exception fall through and changed the error message in the parser to be more informative.

minikube-gradle-plugin/README.md
```diff
@@ -30,11 +37,29 @@ minikubeStart {
...
+
+This plugin also allows users to add in any custom `minikube` and `docker build` task.
 
 ```groovy
 task minikubeCustom(type: com.google.cloud.tools.minikube.MinikubeTask) {
   command = "custom"
   flags = ["--some-flag"]
 }
+
+task minikubeDockerBuildCustom(type: com.google.cloud.tools.minikube.DockerBuildTask) {
```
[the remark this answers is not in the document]
coollog: Removed

minikube-gradle-plugin/README.md
coollog: Added

minikube-gradle-plugin/README.md
```diff
@@ -30,7 +34,20 @@ minikubeStart {
...
-This plugin also allows users to add in any custom minikube task.
+`minikubeDockerBuild` task is of the type `DockerBuildTask` and can be configured with:
+- `context` (`String`) : PATH | URL | - (See 'Extended description' under the [`docker build` Reference](https://docs.docker.com/engine/reference/commandline/build/))
+- `flags` (`String[]`) : any flags to pass to `docker build` (See 'Options' under the [`docker build` Reference](https://docs.docker.com/engine/reference/commandline/build/))
+- `minikube` (`String`) : path to minikube executable which should be set by using the `minikube` extension **users should not edit this for this provided task**
+- `docker` (`String`) : path to Docker executable which should be set by using the `docker` extension **users should not edit this for this provided task**
+
+```groovy
+minikubeDockerBuild {
+  context = "."
```
loosebazooka: I wonder if we can be more helpful here for build, providing reasonable defaults.
1. `dockerfile` param in minikubeDockerBuild which defaults to `src/main/docker/Dockerfile` and configures the task with `-f src/main/docker/Dockerfile`. Since we don't really expect to find Dockerfile in the root of a gradle project.
2. context will probably be in `build/libs/` but there is a way to programmatically extract the final artifact with gradle (not sure if this is what all users want), but we can start with setting the context default to `build/libs`.

This can be in a follow-up PR - and just move what you want into a new issue and start from there.

loosebazooka: Also more context for using src/main/docker <--- seems to be the default for a lot of other gradle plugins (including out gradle plugin).

coollog: Okay I have set the default context to be `build/libs`. I'll file a PR about providing more help for the build.

Issue #14, opened 2017-10-04
Have minikubeDockerBuild be more helpful by providing default parameters.


coollog:
Currently, `minikubeDockerBuild` (#7) takes a [`context` and `flags`](https://docs.docker.com/engine/reference/commandline/build/).

We want it to be more helpful by providing some default `docker build` parameters as part of the task configuration:
 
~~1. `dockerfile`- configures the task with `-f <value>`
    - Default: `src/main/docker/Dockerfile`
    - Reason: since we don't really expect to find Dockerfile in the root of a Gradle project.~~
1. Reason: save the user the hassle of having to define build context/Dockerfile location themselves.
    1. **IN PROGRESS (#16)** A separate task `prepareMinikubeDockerBuild` that prepares the context to use for `minikubeDockerBuild`. Takes parameters:
        - `context`: the task syncs this to `build/docker`
            - Default: `build/libs`
        - `dockerfile`: the task syncs this to `build/docker/Dockerfile`
            - Default: `src/main/docker/Dockerfile`
    1. **IN PROGRESS (#17) ** Have `minikubeDockerBuild` use `build/docker` for its context and depends on the successful execution of the `prepareMinikubeDockerBuild` task.
2. **IN PROGRESS (#15)** `tag` - configures the task with `-t <value>`
    - Default: `${project.group}/${project.name}:${project.version}`
    - Reason: provides the user with a convenient tag to refer to the image with

Please comment with more ideas about what configuration to provide. The above list will be updated with more ideas.

r2d4:
It might be useful to auto-populate the "tag" parameter with something like "$ORGANIZATION/$PROJECT_NAME:$VERISON"

coollog:
Okay, how about having a field like:

`imageTagConfig` - nested with properties: `organization`, `projectName`, `version`
    (Configures the task with `-t <organization>/<projectName>:<version>`)

@loosebazooka

coollog:
Also, @loosebazooka , it seems that the Dockerfile must be in the build context, so a default context of `build/libs` with a default Dockerfile of `src/main/docker/Dockerfile` would not work.

loosebazooka:
I think you can just have those be base properties on the task.

coollog:
What if the user wanted to not use the auto-populated task? Should we include another property called `useAutoGeneratedTag` that defaults to true?

loosebazooka:
[...]

PR #15, opened 2017-10-05
Adds `tag` field to minikubeDockerBuild task with default.

coollog:
Feature 2 of #14 .

### DIFF
--- a/minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/DockerBuildTask.java
+++ b/minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/DockerBuildTask.java
@@ -28,13 +28,16 @@ import org.gradle.api.logging.Logger;
 import org.gradle.api.provider.PropertyState;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.TaskAction;
+import org.gradle.internal.impldep.com.google.common.base.Strings;
 
 /** Task to build Docker images. */
 public class DockerBuildTask extends DefaultTask {
   /**
-   * Name components may contain lowercase letters, digits and separators. A separator is defined as
-   * a period, one or two underscores, or one or more dashes. A name component may not start or end
-   * with a separator.
+   * Matches a name component of a Docker image tag.
+   *
+   * <p>Name components may contain lowercase letters, digits and separators. A separator is defined
+   * as a period, one or two underscores, or one or more dashes. A name component may not start or
+   * end with a separator.
    *
    * @see <a href="https://docs.docker.com/engine/reference/commandline/tag/#description">docker
    *     tag</a>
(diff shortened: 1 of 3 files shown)


minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/DockerBuildTask.java
```diff
@@ -129,9 +142,48 @@ public void execDockerBuild() throws IOException, InterruptedException {
...
+   *
+   * <p>Replaces all instances of ':' in group, name, or version with '.'.
+   *
+   * @return the built tag
+   * @throws GradleException
+   */
+  // @VisibleForTesting
+  String buildDefaultTag() {
+    String group = getProject().getGroup().toString().replace(':', '.');
+    String projectName = getProject().getName().toString().replace(':', '.');
```
loosebazooka: I don't think gradle actually allows colons in projectName, are we expecting : somewhere?

loosebazooka: I wonder if replacing will be confusing, perhaps we just provide validation : "A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes" from https://docs.docker.com/engine/reference/commandline/tag/#description

loosebazooka: We can warn and set the tag to null if we can't generate a reasonable one.

coollog: I tried putting a colon into a project name and it worked. I'll add the warning.

loosebazooka: Oh yeah, huh. Good call.

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/DockerBuildTask.java
```diff
@@ -158,29 +184,47 @@ public void execDockerBuild() throws IOException, InterruptedException {
...
+      return "";
+    }
     tagBuilder.append(projectName);
+
     if (!"unspecified".equals(projectVersion)) {
+      // Checks if project.version can be used as the image tag name.
+      if (!projectVersion.matches(TAG_NAME_REGEX)) {
+        logger.warn(
+                "Default image tag could not be generated because project.version is not a valid tag name");
+        return "";
```
loosebazooka: Wondering if we should return `null` instead of `""` so that the build fails if they don't override it.
Fails because `@Input` on the `DockerBuildTask` wont have a value.

It might look like
```
./gradlew minikubeDockerBuild
WARNING: "blah blah blah, bad tag name"
Task Failed : minikubeDockerBuild because 'tag' was not defined.
```

coollog: Yea that sounds good. Then, if the user truly wanted to not automatically tag the image, they can just set tag to "" explicitly.

coollog:
Closed pending design for new way of building with the Minikube Docker environment.

PR #16, opened 2017-10-06
prepareMinikubeDockerBuild

coollog:
Does 1.i of #14 .


coollog:
Closed pending design for new way of building with the Minikube Docker environment.

Issue #20, opened 2017-10-09
Add MinikubeExtension#getDockerEnv to return the Minikube Docker environment.


coollog:
Add utility function on the `MinikubeExtension`:

`getDockerEnv` -> returns a map from Docker environment variable names to their values

The Docker environment variables are retrieved via the command:

`minikube docker-env --shell=none`

Which gives for example https://github.com/kubernetes/minikube/pull/1792

These variables can then be used to configure a Docker build to build on the Minikube Docker daemon. The user will use some existing Docker build plugin to perform their Docker build.

PR #21, opened 2017-10-09
Removes minikubeDockerBuild.

coollog:
Prepares for #20 .


coollog:
No prob haha, most of the machinery built is still useful.

PR #22, opened 2017-10-09
Adds MinikubeExtension#getDockerEnv utility function.

### LINKED ISSUE #20: Add MinikubeExtension#getDockerEnv to return the Minikube Docker environment.

coollog:
Fixes #20 .

### DIFF
--- a/minikube-gradle-plugin/README.md
+++ b/minikube-gradle-plugin/README.md
@@ -40,3 +40,19 @@ task minikubeCustom(type: com.google.cloud.tools.minikube.MinikubeTask) {
   flags = ["--some-flag"]
 }
 ```
+
+The `minikube` extension also provides a method to obtain the environment variables necessary for working with minikube's docker daemon:
+(See [`docker-machine env` reference](https://docs.docker.com/machine/reference/env/))
+
+```
+def minikubeDockerEnv = minikube.getDockerEnv()
+```
+
+`minikubeDockerEnv` would be a map containing these fields:
+ 
+```
+minikubeDockerEnv.DOCKER_TLS_VERIFY
+minikubeDockerEnv.DOCKER_HOST
+minikubeDockerEnv.DOCKER_CERT_PATH
+minikubeDockerEnv.DOCKER_API_VERSION
+```
--- a/minikube-gradle-plugin/README.md
+++ b/minikube-gradle-plugin/README.md
@@ -48,7 +48,7 @@ The `minikube` extension also provides a method to obtain the environment variab
 def minikubeDockerEnv = minikube.getDockerEnv()
 ```
 
-`minikubeDockerEnv` would be a map containing these fields:
+`minikubeDockerEnv` is a map containing these fields:
  
 ```
 minikubeDockerEnv.DOCKER_TLS_VERIFY
--- a/minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/MinikubeExtension.java
+++ b/minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/MinikubeExtension.java
@@ -19,7 +19,6 @@ package com.google.cloud.tools.minikube;
 import com.google.cloud.tools.minikube.util.CommandExecutor;
 import com.google.cloud.tools.minikube.util.MinikubeDockerEnvParser;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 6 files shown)


minikube-gradle-plugin/README.md
```diff
@@ -40,3 +40,19 @@ task minikubeCustom(type: com.google.cloud.tools.minikube.MinikubeTask) {
...
 ```
+
+The `minikube` extension also provides a method to obtain the environment variables necessary for working with minikube's docker daemon:
+(See [`docker-machine env` reference](https://docs.docker.com/machine/reference/env/))
+
+```
+def minikubeDockerEnv = minikube.getDockerEnv()
+```
+
+`minikubeDockerEnv` would be a map containing these fields:
```
loosebazooka: would be a map -> is a map ?

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/MinikubeExtension.java
```diff
@@ -16,16 +16,42 @@
...
+import org.gradle.api.logging.Logger;
 import org.gradle.api.provider.PropertyState;
 
 /** Minikube configuration extension. */
 public class MinikubeExtension {
   private final PropertyState<String> minikube;
+  private final Logger logger;
+
+  // @VisibleForTesting
+  class CommandExecutorFactory {
```
loosebazooka: Is this class useful outside of minikubeExtension, like in the tasks?

minikube-gradle-plugin/README.md
coollog: Done

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/MinikubeExtension.java
coollog: We could have the `CommandExecutorFactory` be a class until `util`, and `MinikubeTask` could use that instead of calling the `CommandExecutor` constructor.

loosebazooka: Yeah, I mean, I realize it's the clean way to get stuff testable. I was just wondering. Otherwise the extension class itself becomes hard to follow and we really just want to limit it to configuration as much as we can.

coollog: I pushed a commit that moves it to its own class file. Another benefit is that future classes that need to mock the `CommandExecutor` can use the `CommandExecutorFactory` as well.

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/MinikubeExtension.java
```diff
@@ -16,16 +16,32 @@
...
+import java.util.Map;
 import org.gradle.api.Project;
 import org.gradle.api.provider.PropertyState;
 
 /** Minikube configuration extension. */
 public class MinikubeExtension {
   private final PropertyState<String> minikube;
 
+  // @VisibleForTesting
+  MinikubeExtension setCommandExecutorFactory(CommandExecutorFactory commandExecutorFactory) {
```
loosebazooka: Oops, I think you can pass this through the constructor instead of here.

loosebazooka:
```
public MinikubeExtension(Project project, CommandExecutorFactory factory)
```

and then in MinikubePlugin.java when you create the extension, you just pass in a command executor factory, and that same factory can also be set on the Task in MinikubePlugin.java

coollog: So to add it to the MinikubeTask, we would need a MinikubeTask#setCommandExecutorFactory. Is this what we want?

coollog: We would need to add a MinikubeTask#setCommandExecutorFactory method to pass the plugin's command executor factory into MinikubeTask. Is this what we want?

coollog:
asdf

coollog:
asdf

coollog:
Moved the CommandExecutorFactory to be a common one created in the MinikubePlugin. However, one problem with this approach is that the MinikubeTask could be in a state where it has a null commandExecutorFactory.

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/MinikubePlugin.java
```diff
@@ -49,23 +51,28 @@ private void configureMinikubeTaskAdditionCallback() {
             });
```
loosebazooka: I think you can add commandExecutorFactory in here, so it's applied to our and custom user tasks?

coollog: Done

minikube-gradle-plugin/src/main/java/com/google/cloud/tools/minikube/MinikubeExtension.java
```diff
@@ -16,16 +16,30 @@
...
+  private final CommandExecutorFactory commandExecutorFactory;
+
+  public MinikubeExtension(Project project, CommandExecutorFactory commandExecutorFactory) {
     minikube = project.property(String.class);
     setMinikube("minikube");
+
+    this.commandExecutorFactory = commandExecutorFactory;
+  }
+
+  public CommandExecutorFactory getCommandExecutorFactory() {
```
loosebazooka: Do we need this getter?

coollog: Nope, I took it out. This was leftover from another attempt.

loosebazooka:
:ship: ship it

PR #25, opened 2017-11-20
Initialize directory for crepecake.

coollog:


loosebazooka:
Do we want Gradle at the root level?

coollog:
I think since we will build the gradle plugin first, it's probably easier to just have it all be in a gradle project. Then, when we build the maven plugin, we can move the common code into a shared library when we move this codebase to its own separate repo.

Issue #27, opened 2017-11-20
Add SHA-256 digest representation.


coollog:
SHA-256 digest example:

`sha256:41f7343603a17fe746c4bbd5af990b41140edea48be836d0815d57934ca416f5`

PR #29, opened 2017-11-20
Adds SHA-256 hasher.

coollog:
Related to #27

### DIFF
--- a/crepecake/src/main/java/com/google/cloud/tools/crepecake/hash/ByteHasher.java
+++ b/crepecake/src/main/java/com/google/cloud/tools/crepecake/hash/ByteHasher.java
@@ -21,6 +21,7 @@ import java.security.NoSuchAlgorithmException;
 
 /** Static utility class for generating SHA-256 hashes in hexadecimal. */
 public class ByteHasher {
+
   private static final String SHA_256_ALGORITHM = "SHA-256";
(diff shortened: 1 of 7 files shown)


loosebazooka:
I think we want to move away from `util` package naming. If it's a sha hasher, just throw it in `hash` or something like that.

coollog:
Okay, changed the package to `hash`.

crepecake/src/main/java/com/google/cloud/tools/crepecake/hash/ByteHasher.java
```diff
@@ -0,0 +1,42 @@
...
+package com.google.cloud.tools.crepecake.hash;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/** Static utility class for generating SHA-256 hashes in hexadecimal. */
+public class ByteHasher {
+
+  private static final String SHA_256_ALGORITHM = "SHA-256";
+
```
loosebazooka: Can you add a comment to this method?

loosebazooka: oh it looks like it'll just be the same as the class header. Never mind.

PR #35, opened 2017-11-21
Adds JSON parser.

### LINKED ISSUE #30: Add a JSON parser.
Can parse JSON string into Java object, and convert Java object to JSON string.

Should be able to serialize/deserialize digest objects.

Should use GSON library.

coollog:
Fixes #30

### DIFF
--- a/crepecake/build.gradle
+++ b/crepecake/build.gradle
@@ -15,6 +15,7 @@ repositories {
 
 dependencies {
     compile gradleApi()
+    compile 'com.google.code.gson:gson:2.8.2'
     testCompile 'junit:junit:4.12'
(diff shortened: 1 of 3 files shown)


loosebazooka:
I don't think you need to verify that GSON works in the tests. Is there an issue with calling GSON directly?

coollog:
Once I add the Digest serializer, I will be changing these tests to test serializing/deserializing JSON with Digest as a field type. However, it probably does make more sense to just make this be a factory that returns the Gson. Maybe I should just change this to have one static method `getGson()` that returns the constructed Gson?

coollog:
I'll file a followup PR for adding the Digest serializer, and then just add Digests as fields in these tests.

PR #40, opened 2017-11-21
 Adds Digest JSON type adapter.

### LINKED ISSUE #39: Add Digest serializer to JSON parser.

coollog:
Fixes #39


crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonParser.java
```diff
@@ -0,0 +1,41 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.json;
+
+import com.google.cloud.tools.crepecake.image.Digest;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+
+/** Interface to a custom {@link Gson} JSON parser. */
```
loosebazooka: Is this class only used for Digest.class?

crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonParser.java
```diff
@@ -0,0 +1,42 @@
...
+public class JsonParser {
```
loosebazooka: Kinda makes it hard to use this in testing if everything is static? Would it make sense to singleton this or something in that style? Or am I miss-guessing how this will be used?

crepecake/src/test/java/com/google/cloud/tools/crepecake/json/JsonParserTest.java
```diff
@@ -0,0 +1,159 @@
...
+      private int number;
+      private List<String> texts;
+      private List<Digest> digests;
+    }
+  }
+
+  @Test
+  public void testFromJson() throws DigestException {
+    StringBuffer jsonBuffer = new StringBuffer();
+    jsonBuffer.append("{");
```
loosebazooka: Can this be a file in "resources"? It's quite hard to read like this.

crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonParser.java
coollog: Yea, this is sorta a singleton, except that it is a singleton of a Gson class.

coollog:
Changed the JSON parsing library to Jackson since it's faster.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Digest.java
```diff
@@ -0,0 +1,100 @@
...
+    return new Digest(digest, hash);
+  }
+
+  /** Creates a new instance from a valid digest string. */
+  public static Digest fromDigest(String digest) throws DigestException {
+    if (!digest.matches(DIGEST_PATTERN)) {
+      throw new DigestException("Invalid digest: " + digest);
+    }
+
+    Matcher matcher = HashPattern.matcher(digest);
```
elharo: This test seems redundant.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Digest.java
```diff
@@ -0,0 +1,100 @@
...
+  @Override
+  public boolean equals(Object obj) {
+    if (obj instanceof Digest) {
+      return digest.equals(((Digest) obj).digest);
+    }
+
+    return false;
+  }
+
+  private Digest(String digest, String hash) {
```
elharo: Not a rule but we usually put constructors at the top.

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/DigestTest.java
```diff
@@ -0,0 +1,114 @@
...
+      Digest.fromDigest(badDigest);
+      Assert.fail("Invalid digest should have caused digest creation failure.");
+    } catch (DigestException ex) {
+      // pass
+    }
+  }
+
+  @Test
+  public void testUseAsMapKey() throws DigestException {
+    final Digest digestA1 = Digest.fromHash(createGoodHash('a'));
```
elharo: Google rule is to only use final on local variables when mandatory

coollog:
@elharo 

Sorry, I should have made a more explicit note. This PR will essentially be rebased on top of #28 once that is merged, so the `Digest` class code is actually part of that PR. I'll address your comments there.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Digest.java
coollog: Should this go before or after the `public static` methods?

crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonParser.java
```diff
@@ -0,0 +1,61 @@
...
+ * </ul>
+ *
+ * @see <a href="https://github.com/FasterXML/jackson">https://github.com/FasterXML/jackson</a>
+ */
+public class JsonParser {
+
+  private static final ObjectMapper objectMapper;
+
+  static {
+    objectMapper = new ObjectMapper();
```
elharo: you don't seem to need a static block here; just initialize the field

crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonParser.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * {@code @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)}
+ *
+ * <ul>
+ *   <li> Fields that are private are also accessible for serialization/deserialization.
+ * </ul>
+ *
+ * @see <a href="https://github.com/FasterXML/jackson">https://github.com/FasterXML/jackson</a>
+ */
+public class JsonParser {
```
elharo: This seems to be a serializer, not a parser.

coollog: This will also have deserialization methods. Would JsonHelper be a better name?

crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonTemplate.java
```diff
@@ -14,25 +14,16 @@
...
-  public static String toJson(Object src) {
-    return GSON.toJson(src);
-  }
-
-  public static <T> T fromJson(String json, Class<T> classOfT) {
-    return GSON.fromJson(json, classOfT);
-  }
-}
+/**
+ * All JSON templates to be used with {@link JsonParser} must extends this class.
```
elharo: must extend

crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonParser.java
elharo: probably

crepecake/src/test/java/com/google/cloud/tools/crepecake/json/JsonParserTest.java
```diff
@@ -0,0 +1,86 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.json;
+
+import com.google.cloud.tools.crepecake.image.DescriptorDigest;
+import java.io.*;
```
elharo: no wildcard imports

crepecake/src/test/resources/json/basic.json
```diff
@@ -0,0 +1 @@
+{"number":54,"text":"crepecake","digest":"sha256:8c662931926fa990b41da3c9f42663a537ccd498130030f9149173a0493832ad","innerObject":{"number":23,"texts":["first text","second text"],"digests":["sha256:91e0cae00b86c289b33fee303a807ae72dd9f0315c16b74e6ab0cdbe9d996c10","sha256:4945ba5011739b0b98c4a41afe224e417f47c7c99b2ce76830999c9a0861b236"]},"list":[{"number":42,"texts":[]},{"number":99,"texts":["some text"],"digests":["sha256:d38f571aa1c11e3d516e0ef7e513e7308ccbeb869770cb8c4319d63b10a0075e"]}]}
```
elharo: would be nice to format this more legibly with spaces and indents and line breaks

coollog: The main problem with formatting this is that it won't be able to match the output of `JsonHelper#writeJson` as that will stream the JSON in the minified form. The solution would be

- to add an option to the `writeJson` method to be able to pretty print or 
- have the test minify the basic.json after reading it
- leave it as is, since the file can be read legibly with any pretty JSON reader

LYMWYT

crepecake/src/test/java/com/google/cloud/tools/crepecake/json/JsonHelperTest.java
```diff
@@ -0,0 +1,90 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.URISyntaxException;
+import java.security.DigestException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import org.gradle.internal.impldep.com.google.common.io.CharStreams;
```
elharo: This looks like the wrong import. Should probably be c.g.common.io.CharStreams

crepecake/src/test/java/com/google/cloud/tools/crepecake/json/JsonHelperTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+  private static class TestJson extends JsonTemplate {
+    private int number;
+    private String text;
+    private DescriptorDigest digest;
+    private InnerObject innerObject;
+    private List<InnerObject> list;
+
+    private static class InnerObject extends JsonTemplate {
+      private int number;
```
elharo: this field shadows the above field. Rename to avoid confusion.

crepecake/src/test/resources/json/basic.json
elharo: Ideally we wouldn't fail the test on differences in insignificant white space -- i.e. use a smarter comparison function -- but for now this will work.

crepecake/src/test/java/com/google/cloud/tools/crepecake/json/JsonHelperTest.java
coollog: I named this the same intentionally to check an inner JSON object with a field having the same name works fine.

elharo: In that case, add a comment making this explicit.

Issue #34, opened 2017-11-20
Add a tar.gzip stream builder.


coollog:
`.addEntry(File file, String path)`
Adds the `file` to the tar.gzip stream with the extraction path `path`.

Issue #33, opened 2017-11-20
Add a layer representation.


coollog:
Should consist of `digest`, `size`, and `diffId`.

If the layer is cached, all of these values should be able to be loaded directly.
If the layer is not cached (or needs to be pushed to the registry), it should have another field `content` that is the actual tar.gzip layer blob. In the case that the layer is pulled from a registry, it might not have `diffId` and that would need to be computed by uncompressing the layer blob.

PR #42, opened 2017-11-21
Adds BlobStream to hold BLOBs.

coollog:
Represent the BLOBs used to store and send image content (layer tar.gzip, manifest JSON, container configuration JSON).

Required for #33 and #34 .


crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
```diff
@@ -31,7 +31,7 @@ public BlobStream() {
     byteArrayOutputStream = new ByteArrayOutputStream(0);
   }
 
-  /** Initializes with the contents of the input blob. */
+  /** Initializes with the contents of the {@param inputStream}. */
```
elharo: I don't think this is how you use @param in Javadoc.

coollog: Is there a correct way to refer to the name of the parameter or should I just say `{@code inputStream} parameter`?

elharo: @code or @link, I think

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
```diff
@@ -0,0 +1,50 @@
...
+package com.google.cloud.tools.crepecake.blob;
+
+import com.google.common.io.ByteStreams;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/** A read-only {@link OutputStream} for BLOBs. */
+public class BlobStream {
```
loosebazooka: Can you link to the PR where this is used? It just seems like too simple a wrapper in this form.

coollog: Yeah, it is used in #43 #44 and #45 currently.

The idea is to have this be the common datatype for passing around BLOBs that can be instantiated from various ways:

- from an InputStream - reading from a file or HTTP response
- from a String - from a JSON string
- from another OutputStream - the built layer tar stream

Is this sufficient for having this wrapper?

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
```diff
@@ -0,0 +1,50 @@
...
+
+  /** Initializes an empty BLOB. */
+  public BlobStream() {
+    byteArrayOutputStream = new ByteArrayOutputStream(0);
+  }
+
+  /** Initializes with the contents of the {@code inputStream} parameter. */
+  public BlobStream(InputStream inputStream) throws IOException {
+    byteArrayOutputStream = new ByteArrayOutputStream();
+    ByteStreams.copy(inputStream, byteArrayOutputStream);
```
elharo: as a general rule, this much work in a constructor is iffy. Constructors should usually be pretty fast. We often use some sort of static factory method if we need to do something like this.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
```diff
@@ -0,0 +1,50 @@
...
+
+  /** Initializes with a string. */
+  public BlobStream(String content) throws IOException {
+    final byte[] contentBytes = content.getBytes();
+    byteArrayOutputStream = new ByteArrayOutputStream(contentBytes.length);
+    byteArrayOutputStream.write(contentBytes);
+  }
+
+  public void writeTo(OutputStream outputStream) throws IOException {
+    byteArrayOutputStream.writeTo(outputStream);
```
elharo: This can only be done once. Might make more sense to store the byte array or String (which you're doing anyway, just implicitly) so you don't have to worry about that.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
elharo: This class is not a stream so please don't name it that.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
```diff
@@ -0,0 +1,50 @@
...
+  /** Initializes with the contents of the {@code inputStream} parameter. */
+  public BlobStream(InputStream inputStream) throws IOException {
+    byteArrayOutputStream = new ByteArrayOutputStream();
+    ByteStreams.copy(inputStream, byteArrayOutputStream);
+  }
+
+  /** Initializes with a string. */
+  public BlobStream(String content) throws IOException {
+    final byte[] contentBytes = content.getBytes();
+    byteArrayOutputStream = new ByteArrayOutputStream(contentBytes.length);
```
elharo: need to specify encoding here, "UTF-8"

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
elharo: I'm vert nervous about the design here. I think you should probably store a byte array instead, and convert it to a ByteArrayOutputStream as needed.

crepecake/src/test/java/com/google/cloud/tools/crepecake/blob/BlobStreamTest.java
```diff
@@ -0,0 +1,66 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.blob;
+
+import java.io.*;
```
elharo: no wildcard imports in Google style

coollog:
@elharo Changing this to be an actual stream

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/AbstractHashingBlobStream.java
```diff
@@ -0,0 +1,55 @@
...
+package com.google.cloud.tools.crepecake.blob;
+
+import com.google.cloud.tools.crepecake.hash.ByteHashBuilder;
+import com.google.cloud.tools.crepecake.image.DescriptorDigest;
+import com.google.cloud.tools.crepecake.image.DigestException;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.security.NoSuchAlgorithmException;
+
+abstract class AbstractHashingBlobStream implements BlobStream {
```
elharo: Do you need an interface AND an abstract class? Are there multiple subclasses of each?

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/EmptyBlobStream.java
```diff
@@ -0,0 +1,28 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.blob;
+
+import com.google.cloud.tools.crepecake.hash.ByteHashBuilder;
+import java.io.IOException;
+import java.io.OutputStream;
+
+public class EmptyBlobStream extends AbstractHashingBlobStream {
```
elharo: What's the purpose of this class? Javadoc might help here.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/ProvidedFileBlobStream.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.cloud.tools.crepecake.hash.ByteHashBuilder;
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/** A {@link BlobStream} that streams from a {@link File}. */
+class ProvidedFileBlobStream extends AbstractHashingBlobStream {
```
elharo: Just FileBlobStream should be enough

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/ProvidedInputStreamBlobStream.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.google.cloud.tools.crepecake.blob;
+
+import com.google.cloud.tools.crepecake.hash.ByteHashBuilder;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/** A {@link BlobStream} that streams from an {@link InputStream}. */
+class ProvidedInputStreamBlobStream extends AbstractHashingBlobStream {
```
elharo: InputStreamBlobStream

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/ProvidedFileBlobStream.java
```diff
@@ -0,0 +1,50 @@
...
+  private final byte[] byteBuffer = new byte[8192];
+
+  ProvidedFileBlobStream(File file) {
+    this.file = file;
+  }
+
+  @Override
+  protected void writeToAndHash(OutputStream outputStream, ByteHashBuilder byteHashBuilder)
+      throws IOException {
+    InputStream fileStream = new BufferedInputStream(new FileInputStream(file));
```
elharo: Use try with resources here to ensure the file is closed when you're done or an exception is thrown

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/ProvidedInputStreamBlobStream.java
```diff
@@ -0,0 +1,45 @@
...
+    this.inputStream = inputStream;
+  }
+
+  @Override
+  protected void writeToAndHash(OutputStream outputStream, ByteHashBuilder byteHashBuilder)
+      throws IOException {
+    int bytesRead;
+    while ((bytesRead = inputStream.read(byteBuffer)) != -1) {
+      // Writes to the output stream and builds the BLOB's hash as well.
+      outputStream.write(byteBuffer, 0, bytesRead);
```
elharo: outputStream.flush for safety

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/ProvidedStringBlobStream.java
```diff
@@ -0,0 +1,38 @@
...
+  private final byte[] contentBytes;
+
+  ProvidedStringBlobStream(String content) {
+    contentBytes = content.getBytes(Charsets.UTF_8);
+  }
+
+  @Override
+  protected void writeToAndHash(OutputStream outputStream, ByteHashBuilder byteHashBuilder)
+      throws IOException {
+    outputStream.write(contentBytes);
```
elharo: outputStream.flush

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/ProvidedWriterBlobStream.java
```diff
@@ -0,0 +1,38 @@
...
+  private final BlobStreamWriter writer;
+
+  ProvidedWriterBlobStream(BlobStreamWriter writer) {
+    this.writer = writer;
+  }
+
+  @Override
+  protected void writeToAndHash(OutputStream outputStream, ByteHashBuilder byteHashBuilder)
+      throws IOException {
+    writer.writeTo(outputStream);
```
elharo: flush

crepecake/src/main/java/com/google/cloud/tools/crepecake/hash/ByteHashBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * Generates SHA-256 hashes in hexadecimal. Can act as an {@link OutputStream} that captures the
+ * bytes to hash.
+ */
+public class ByteHashBuilder extends OutputStream {
```
elharo: You may be reinventing java.security.DigestOutputStream which implements SHA-256

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
```diff
@@ -0,0 +1,47 @@
...
+
+/** A stream for BLOBs. */
+public interface BlobStream {
+
+  /**
+   * Writes the BLOB to an {@link OutputStream}.
+   *
+   * @param outputStream the {@link OutputStream} to write to
+   */
+  void writeTo(OutputStream outputStream)
```
elharo: It's unclear why this general interface should throw NoSuchAlgorithmException or DigestException. I'm not sure it does throw NoSuchAlgorithmException. If DigestException is needed, then an @param is called for to explain it.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
```diff
@@ -0,0 +1,47 @@
...
+      throws IOException, NoSuchAlgorithmException, DigestException;
+
+  /**
+   * This is only valid <b>after</b> {@code writeTo} is called.
+   *
+   * @return the {@link BlobDescriptor} of the written BLOB
+   * @throws IllegalStateException if {@code writeTo} has not been called
+   */
+  @Nonnull
+  default BlobDescriptor getWrittenBlobDescriptor() throws IllegalStateException {
```
elharo: Is there such a thing as an UnwrittenBlobDescriptor? If not, maybe just getBlobDescriptor.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobDescriptor.java
```diff
@@ -0,0 +1,68 @@
...
+  public DescriptorDigest getDigest() {
+    return digest;
+  }
+
+  public long getSize() {
+    return size;
+  }
+
+  /**
+   * Two {@link BlobDescriptor} objects are equal if their
```
elharo: Should two BlobDescriptors be equal if they have the same size and both digests are null?

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobDescriptor.java
```diff
@@ -0,0 +1,68 @@
...
+   *
+   * <ol>
+   *   <li>{@code digest}s are not null and equal
+   *   <li>{@code size} is equal
+   * </ol>
+   */
+  @Override
+  public boolean equals(Object obj) {
```
elharo: when you override equals, you need to override hashCode too

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobDescriptor.java
coollog: They shouldn't because the BlobDescriptor is foremost equal by digest. Without digests, BlobDescriptors cannot be deemed equal by size alone. The only use case for BlobDescriptors without digest is the image manifest, which should not be compared for equality by just their size alone.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
coollog: The reasoning behind this naming was to make it more clear that this is the BlobDescriptor after `writeTo` is called. I felt that just `getBlobDescriptor` would imply that the BlobDescriptor would be available even before `writeTo` is called.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
coollog: The NoSuchAlgorithmException is because the writing is mostly done through CountingDigestOutputStream, which performs a SHA-256 hash as the output is written. I wrapped this in IOException to make this interface more clear. 

DigestException seems to still fit this interface since getWrittenBlobDescriptor is getting the digest. I added an explanation for DigestException.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStream.java
coollog: Actually, I should just make this be returned from writeTo..

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/AbstractHashingBlobStream.java
```diff
@@ -38,7 +35,8 @@
   protected abstract void writeToAndHash(OutputStream outputStream) throws IOException;
 
   @Override
-  public void writeTo(OutputStream outputStream) throws IOException, DigestException {
+  public final BlobDescriptor writeTo(OutputStream outputStream)
+      throws IOException, DigestException {
     CountingDigestOutputStream hashingOutputStream;
```
elharo: move the declaration into the try block where you initialize this

crepecake/src/main/java/com/google/cloud/tools/crepecake/hash/CountingDigestOutputStream.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static final String SHA_256_ALGORITHM = "SHA-256";
+
+  /** Keeps track of the total number of bytes appended. */
+  private long totalBytes = 0;
+
+  /** Wraps the {@code outputStream}. */
+  public CountingDigestOutputStream(OutputStream outputStream) throws NoSuchAlgorithmException {
+    super(outputStream, null);
+    setMessageDigest(MessageDigest.getInstance(SHA_256_ALGORITHM));
```
elharo: Peer Javadoc:

Every implementation of the Java platform is required to support the following standard MessageDigest algorithms:

MD5
SHA-1
SHA-256

Given that, I think it would be reasonable to to catch the NoSuchAlgorithmException and convert it into a RuntimeException here, since failure to support this algorithm indicates a broken VM.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/AbstractHashingBlobStream.java
```diff
@@ -0,0 +1,52 @@
...
+   * @param outputStream the {@link DigestOutputStream} to write to
+   */
+  protected abstract void writeToAndHash(OutputStream outputStream) throws IOException;
+
+  @Override
+  public final BlobDescriptor writeTo(OutputStream outputStream)
+      throws IOException, DigestException {
+    CountingDigestOutputStream hashingOutputStream;
+    try {
+      hashingOutputStream = new CountingDigestOutputStream(outputStream);
```
elharo: I'm not sure you need the catch block here but if you do, all the code should be inside the try block.

PR #43, opened 2017-11-21
Adds image layer representation.

### LINKED ISSUE #33: Add a layer representation.

coollog:
Fixes #33


coollog:
Hold off on review for this. Will be changing changing how this should work.

coollog:
Okay, made this just adding a layer representation.

coollog:
This is good for review now.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/CachedLayer.java
```diff
@@ -0,0 +1,59 @@
...
+ * A {@link Layer} that has been written out (i.e. to a cache) and has its file-backed content BLOB,
+ * digest, size, and diff ID.
+ */
+public class CachedLayer extends Layer {
+
+  private final File file;
+  private final BlobDescriptor blobDescriptor;
+  private final DescriptorDigest diffId;
+
+  public CachedLayer(File file, BlobDescriptor blobDescriptor, DescriptorDigest diffId) {
```
elharo: diffId? this could use some Javadoc to explain

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/CachedLayer.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.cloud.tools.crepecake.blob.BlobStreams;
+import java.io.File;
+
+/**
+ * A {@link Layer} that has been written out (i.e. to a cache) and has its file-backed content BLOB,
+ * digest, size, and diff ID.
+ */
+public class CachedLayer extends Layer {
+
+  private final File file;
```
elharo: How are these related? any file with any descriptor and any digest? or do they have to match? If they have to match somehow, will we detect it here if they don't?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/CachedLayer.java
```diff
@@ -0,0 +1,59 @@
...
+    return BlobStreams.from(file);
+  }
+
+  @Override
+  public LayerType getType() {
+    return LayerType.CACHED;
+  }
+
+  @Override
+  public BlobDescriptor getBlobDescriptor() {
```
elharo: These types are immutable, right?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Layer.java
```diff
@@ -0,0 +1,72 @@
...
+  public abstract LayerType getType();
+
+  /**
+   * @return the layer's content {@link BlobDescriptor}
+   * @throws LayerException if not available
+   */
+  public abstract BlobDescriptor getBlobDescriptor() throws LayerException;
+
+  /**
+   * @return the layer's diff ID
```
elharo: this doesn't really tell me what a diff ID is though

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/LayerException.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.image;
+
+/** {@link Exception} thrown when accessing non-existent properties of layers. */
```
elharo: no need to link to Exception

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/LayerType.java
```diff
@@ -0,0 +1,31 @@
+package com.google.cloud.tools.crepecake.image;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/CachedLayer.java
coollog: When caching, the UnwrittenLayer is written out as its BLOB, descriptor, and diffId. There will be no checks at runtime in CachedLayer for matching and the user of CachedLayer must ensure that the descriptor and diffId matches the BLOB stored in the file. I will add a note about this in the javadoc.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/CachedLayer.java
coollog: Yes, BlobDescriptor is immutable.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Layer.java
coollog: I have the description above in the class javadoc - should I duplicate that to here as well?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/LayerType.java
coollog: Added copyright.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Layer.java
```diff
@@ -0,0 +1,72 @@
...
+ *   <li>Diff ID
+ *   <li>
+ *       <ul>
+ *         <li>The SHA-256 hash of the uncompressed archive (tarball) of the partial filesystem
+ *             changeset.
+ *       </ul>
+ *
+ * </ul>
+ */
+public abstract class Layer {
```
loosebazooka: Why does this need to be an abstract class?

loosebazooka: and not an interface?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/LayerException.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.image;
+
+/** Exception thrown when accessing non-existent properties of layers. */
```
loosebazooka: Is this only thrown for Layer properties that don't exist? Can we call it something like that?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/CachedLayer.java
```diff
@@ -0,0 +1,69 @@
...
+
+package com.google.cloud.tools.crepecake.image;
+
+import com.google.cloud.tools.crepecake.blob.BlobDescriptor;
+import com.google.cloud.tools.crepecake.blob.BlobStream;
+import com.google.cloud.tools.crepecake.blob.BlobStreams;
+import java.io.File;
+
+/**
+ * A {@link Layer} that has been written out (i.e. to a cache) and has its file-backed content BLOB,
```
elharo: written out to a cache

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/CachedLayer.java
elharo: Could the constructor be non-public to make it harder to create a layer where these pieces don't match? Or could we reorganize the API so they always match?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Layer.java
elharo: No, that's good. Thanks.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/ReferenceLayer.java
```diff
@@ -0,0 +1,53 @@
...
+public class ReferenceLayer extends Layer {
+
+  /** The {@link BlobDescriptor} of the compressed layer content. */
+  private final BlobDescriptor blobDescriptor;
+
+  /** The digest of the uncompressed layer content. */
+  private final DescriptorDigest diffId;
+
+  /** Instantiate with a {@link BlobDescriptor} and diff ID. */
+  public ReferenceLayer(BlobDescriptor blobDescriptor, DescriptorDigest diffId) {
```
elharo: Why is the DescriptorDigest independent of the BlobDescriptor instead of part of it? Could a single Blob have different descriptors?

coollog: A layer BLOB has its main DescriptorDigest that is a hash of the compressed layer tarball archive, but it also has a DescriptorDigest diff ID that is a hash of the uncompressed layer tarball archive.

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/UnwrittenLayerTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+/** Tests for {@link UnwrittenLayer}. */
+public class UnwrittenLayerTest {
+
+  @Rule public TemporaryFolder fakeFolder = new TemporaryFolder();
+
+  @Mock private Blob mockCompressedBlob;
+
```
elharo: nit: probably no blank lines in between these Mock fields

coollog:
@loosebazooka requesting a final pass

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/AbstractHashingBlob.java
```diff
@@ -20,27 +20,28 @@
...
 
-    writeToAndHash(hashingOutputStream);
+    writeToWithHashing(hashingOutputStream);
     hashingOutputStream.flush();
 
-    return hashingOutputStream.toBlobDescriptor();
+    try {
+      return hashingOutputStream.toBlobDescriptor();
+    } catch (DigestException ex) {
+      throw new IOException("BLOB hashing failed: " + ex.getMessage());
```
elharo: include the nested exception

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/AbstractHashingBlob.java
```diff
@@ -20,27 +20,28 @@
...
-   * @param outputStream the {@link DigestOutputStream} to write to
+   * @param outputStream the {@link CountingDigestOutputStream} to write to
    */
-  protected abstract void writeToAndHash(OutputStream outputStream) throws IOException;
+  abstract void writeToWithHashing(CountingDigestOutputStream outputStream) throws IOException;
 
   @Override
-  public final BlobDescriptor writeTo(OutputStream outputStream)
-      throws IOException, DigestException {
+  public final BlobDescriptor writeTo(OutputStream outputStream) throws IOException {
```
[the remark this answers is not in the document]
coollog: Okay, let me open a separate issue for that.

PR #44, opened 2017-11-21
Adds a tar.gzip stream builder.

### LINKED ISSUE #34: Add a tar.gzip stream builder.

coollog:
Fixes #34


crepecake/src/test/java/com/google/cloud/tools/crepecake/tar/TarStreamBuilderTest.java
```diff
@@ -0,0 +1,69 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.tar;
+
+import com.google.cloud.tools.crepecake.blob.BlobStream;
+import com.google.common.io.CharStreams;
+import java.io.*;
```
elharo: Google style is not to use wildcard imports.

loosebazooka: Hah, it's just a placeholder.

loosebazooka:
Hah, @elharo naming is just a placeholder.

coollog:
Yea, @elharo the naming is temporary. But, it refers to the thin layers of this image building.

elharo:
What do expect the package name to be? Repo history is cleaner if we don't have to rename packages.

coollog:
So, this repo (minikube-build-tools-for-java) is only a temporary place to put the code for this project right now and start implementing some parts. Once a permanent separate repo is established, all the code here would be moved to the permanent repo with the `crepecake` package renamed.

crepecake/src/main/java/com/google/cloud/tools/crepecake/tar/TarStreamBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+   *     another {@link OutputStream}. This wrap can be a compressor.
+   * @return a {@link BlobStream} containing the built archive BLOB.
+   */
+  private BlobStream toBlobStream(Function<OutputStream, OutputStream> wrapOutputStreamFunction)
+      throws IOException {
+    // Creates an underlying byte stream.
+    ByteArrayOutputStream tarByteStream = new ByteArrayOutputStream();
+
+    // Wraps the underlying byte stream with the provided function.
+    OutputStream wrappedOutputStream;
```
elharo: use try with resources

crepecake/src/main/java/com/google/cloud/tools/crepecake/tar/TarStreamBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+    try {
+      wrappedOutputStream = wrapOutputStreamFunction.apply(tarByteStream);
+    } catch (RuntimeException ex) {
```
elharo: why do you expect a runtime exception to be thrown here? which one?

crepecake/src/main/java/com/google/cloud/tools/crepecake/tar/TarStreamBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+    // Writes each entry in the filesystem to the tarball archive stream.
+    TarArchiveOutputStream tarArchiveOutputStream = new TarArchiveOutputStream(wrappedOutputStream);
+
+    for (Entry entry : entries) {
+      tarArchiveOutputStream.putArchiveEntry(entry.header);
+      ByteStreams.copy(entry.content, tarArchiveOutputStream);
+      tarArchiveOutputStream.closeArchiveEntry();
+    }
+
+    tarArchiveOutputStream.close();
```
elharo: could autoclose this with try-with-resources

crepecake/src/main/java/com/google/cloud/tools/crepecake/tar/TarStreamBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+
+    return new BlobStream(tarByteStream);
+  }
+
+  private OutputStream applyCompressor(OutputStream outputStream) {
+    try {
+      return new CompressorStreamFactory()
+          .createCompressorOutputStream(CompressorStreamFactory.GZIP, outputStream);
+    } catch (CompressorException ex) {
```
elharo: probably just bubble this exception; no need to wrap in a runtimeexception. Handle it further up the call stack.

crepecake/src/test/java/com/google/cloud/tools/crepecake/blob/BlobStreamTest.java
```diff
@@ -0,0 +1,66 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.blob;
+
+import java.io.*;
```
elharo: no wild card imports

crepecake/src/test/java/com/google/cloud/tools/crepecake/blob/BlobStreamTest.java
```diff
@@ -0,0 +1,66 @@
...
+    String output = outputStream.toString();
+
+    Assert.assertEquals("", output);
+  }
+
+  @Test
+  public void testFromInputStream() throws IOException {
+    String expected = "crepecake";
+
+    InputStream inputStream = new ByteArrayInputStream(expected.getBytes());
```
elharo: encoding

crepecake/src/test/java/com/google/cloud/tools/crepecake/tar/TarStreamBuilderTest.java
```diff
@@ -0,0 +1,115 @@
...
+public class TarStreamBuilderTest {
+
+  @Test
+  public void testBuild_compressed() throws IOException, URISyntaxException, CompressorException {
+    testBuild(
+        tarStreamBuilder -> {
+          try {
+            return tarStreamBuilder.toBlobStreamCompressed();
+          } catch (IOException ex) {
+            throw new RuntimeException(ex);
```
elharo: shouldn't need to wrap these in runtime exceptions. If that's required for lambdas, then lambdas are inappropriate here.

crepecake/src/test/java/com/google/cloud/tools/crepecake/tar/TarStreamBuilderTest.java
```diff
@@ -0,0 +1,115 @@
...
+          }
+        },
+        inputStream -> {
+          try {
+            return new CompressorStreamFactory().createCompressorInputStream(inputStream);
+          } catch (CompressorException ex) {
+            throw new RuntimeException(ex);
```
elharo: shouldn't need to wrap these in runtime exceptions

coollog:
Removed the compression option in `TarStreamBuilder` as the preferred workflow should be:

1. Cache the uncompressed tarball archive.
2. Send the tarball archive over HTTP while wrapping the output with a compressor.

coollog:
@loosebazooka requesting a final pass

crepecake/src/main/java/com/google/cloud/tools/crepecake/tar/TarStreamBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
+
+/** Builds a tarball archive. */
+public class TarStreamBuilder {
```
loosebazooka: I wonder if this class should provide a stronger contract to its internals. Since we're delegating to "writeEntriesAsTarArchive" kinda lazily, is it possible that we call addFile after toBlob and change the expectation of what's written in toBlob?

Would it make sense to have a TarBlobBuilder that returns an immutable TarBlob?

coollog: I changed this to make it be locked-in when calling toBlob.

PR #45, opened 2017-11-22
Adds HTTP request utilities.

### LINKED ISSUE #38: Add utilities HttpRequest and HttpResponse for sending HTTP requests and processing the output.

coollog:
Fixes #38


crepecake/src/main/java/com/google/cloud/tools/crepecake/http/ConnectionFactory.java
```diff
@@ -0,0 +1,29 @@
...
+package com.google.cloud.tools.crepecake.http;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+class ConnectionFactory {
+
+  /** Opens a connection to a URL. */
+  public HttpURLConnection newConnection(URL url) throws IOException {
```
elharo: doesn't need to be public

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/ConnectionFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.http;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+class ConnectionFactory {
```
elharo: I'm not sure this class does enough to be worth its weight.

coollog: This class is only for testing purposes since URL is final and can't be mocked.

coollog:
Hmm, yea I'll probably just switch to using the Google HTTP client library. I should probably change the JSON parser to use Jackson instead of GSON since the library recommends it as the faster option.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Request.java
```diff
@@ -0,0 +1,89 @@
...
+import java.net.URL;
+import javax.annotation.Nullable;
+
+/** Sends an HTTP request. */
+public class Request {
+
+  private static final HttpRequestFactory HTTP_REQUEST_FACTORY =
+      new NetHttpTransport().createRequestFactory();
+
+  private static class HttpMethod {
```
elharo: I'm surprised these aren't already constants in some library we already import.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Request.java
```diff
@@ -0,0 +1,89 @@
...
+  public Response send() throws IOException {
+    return send(new BlobStream());
+  }
+
+  public Request setContentType(String contentType) {
+    headers.setContentType(contentType);
+    return this;
+  }
+
+  public Request setMethodPut() {
```
elharo: This is a bit weird. What of instead there were put/post/get methods instead of send?

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Request.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @VisibleForTesting
+  Request(URL url, HttpHeaders headers) throws IOException {
+    this(url);
+    this.headers = headers;
+  }
+
+  /** Sends request with body. */
+  public Response send(BlobStream body) throws IOException {
```
elharo: How big are these? As BlobStream is currently designed, this might as well be a byte array. Should we consider true streaming instead?

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.api.client.http.HttpRequest;
+import com.google.api.client.http.HttpResponse;
+import com.google.cloud.tools.crepecake.blob.BlobStream;
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/** Lazily captures an HTTP response. */
+public class Response implements Closeable {
```
elharo: Does this have to work in multiple threads?

crepecake/src/test/java/com/google/cloud/tools/crepecake/blob/BlobStreamTest.java
```diff
@@ -0,0 +1,66 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.blob;
+
+import java.io.*;
```
elharo: no wildcard imports

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Request.java
coollog: As discussed offline, this will involve a change to BlobStream in #42

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
coollog: So this could be passed between threads, but the reading of the response InputStream should only be done in one thread.

coollog:
Merged in master - ready for review again.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStreamHttpContent.java
```diff
@@ -0,0 +1,42 @@
+package com.google.cloud.tools.crepecake.blob;
```
elharo: copyright

coollog: Added copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobStreamHttpContent.java
```diff
@@ -0,0 +1,58 @@
...
+
+package com.google.cloud.tools.crepecake.blob;
+
+import com.google.api.client.http.HttpContent;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.security.DigestException;
+
+/** {@link Blob}-backed {@link HttpContent}. */
+class BlobStreamHttpContent implements HttpContent {
```
elharo: BlobHttpContent

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobHttpContent.java
```diff
@@ -0,0 +1,58 @@
...
+
+package com.google.cloud.tools.crepecake.blob;
+
+import com.google.api.client.http.HttpContent;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.security.DigestException;
+
+/** {@link Blob}-backed {@link HttpContent}. */
+class BlobHttpContent implements HttpContent {
```
loosebazooka: Can you explain why this is necessary in the javadoc? I'm assuming it's all about keeping the blob content out of memory?

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/Blobs.java
```diff
@@ -41,4 +42,9 @@ public static Blob from(String content, boolean hashing) {
   public static Blob from(BlobWriter writer) {
     return new HashingWriterBlob(writer);
   }
+
+  /** Wraps a {@link Blob} in an {@link HttpContent}. */
+  public static HttpContent toHttpContent(Blob blob) {
```
loosebazooka: This doesn't appear to be an implementation or specific implementation of Blob strictly speaking. Should we limit Blobs to purely the creation of new Blob objects?

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobHttpContent.java
coollog: This is an adapter from Blob to HttpContent, which is what is used by the Google HTTP client. This is so that our `Request` class can directly use Blob.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Request.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.cloud.tools.crepecake.blob.Blob;
+import com.google.cloud.tools.crepecake.blob.Blobs;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.net.URL;
+import javax.annotation.Nullable;
+
+/** Sends an HTTP request. */
+public class Request {
```
loosebazooka: Can you link to a PR where request is used?

coollog: I haven't put up any PRs where request is used, but the relevant issue is: https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/59

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/Blobs.java
coollog: Yes, I moved this to the http package instead.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.api.client.http.HttpRequest;
+import com.google.api.client.http.HttpResponse;
+import com.google.cloud.tools.crepecake.blob.Blob;
+import com.google.cloud.tools.crepecake.blob.Blobs;
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/** Lazily captures an HTTP response. */
```
elharo: What's lazy about this?

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
```diff
@@ -0,0 +1,77 @@
...
+public class Response implements Closeable {
+
+  private final HttpRequest request;
+
+  @Nullable private HttpResponse response;
```
elharo: This confused me. We have a response field that is not a Response. This class might need to be renamed. Maybe HttpConnection? Is anyone else using that name?

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
coollog: Maybe a renaming of the two classes would make sense?

http.Request -> http.HttpMessage
http.Response -> http.HttpResponseReceiver

coollog: Since Request is building the HTTP message and Response is a receiver for the HTTP response.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/InputStreamBlob.java
```diff
@@ -27,6 +27,7 @@
 
   private final byte[] byteBuffer = new byte[8192];
 
+  /** The caller should manage the closing of {@code inputStream}. */
```
elharo: manage the closing of --> close

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Request.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /** Sets the {@code Content-Type} header. */
+  public Request setContentType(String contentType) {
+    headers.setContentType(contentType);
+    return this;
+  }
+
+  /** Sends the request with method GET. */
+  public Response get() throws IOException {
```
elharo: This design keeps surprising me. I would expect the get() or post() methods to make the HTTP connection. Instead the call is lazy, not made until we read from the response.
What if we took a more conventional approach in which Request and Response were immutable, non-lazy holder objects, and a third Http class provided get and post methods that connected?

coollog:
Changed to 3 classes:

1. Request - builds the request
2. Connection - sends the request
3. Response - holds the response

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Connection.java
```diff
@@ -0,0 +1,96 @@
...
+  public Response get(Request request) throws IOException {
+    return send(request, (url, body) -> requestFactory.buildGetRequest(url));
+  }
+
+  /** Sends the request with method POST. */
+  public Response post(Request request) throws IOException {
+    return send(request, requestFactory::buildPostRequest);
+  }
+
+  /** Sends the request with method PUT. */
```
elharo: do we actually need PUT?

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Connection.java
```diff
@@ -0,0 +1,96 @@
...
+  public void close() throws IOException {
+    if (httpResponse == null) {
+      return;
+    }
+
+    httpResponse.disconnect();
+  }
+
+  /** Sends the request with method GET. */
+  public Response get(Request request) throws IOException {
```
elharo: Given that the request factory is essentially a constant, it feels like this can be much simplified with static get and post methods that take a request and a URL.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Connection.java
```diff
@@ -0,0 +1,96 @@
...
+  public Response post(Request request) throws IOException {
+    return send(request, requestFactory::buildPostRequest);
+  }
+
+  /** Sends the request with method PUT. */
+  public Response put(Request request) throws IOException {
+    return send(request, requestFactory::buildPutRequest);
+  }
+
+  /** Function to build an {@link HttpRequest} from a url and a body. */
```
elharo: The functional interface obscures whats really happening. It took me quite a few minutes to decode this in my head. I think it would be simpler to implement get, post, and (if you need it) put in each method. There's not a lot to be gained by sharing the minimal amount of code in `send()`.

crepecake/src/test/java/com/google/cloud/tools/crepecake/http/ResponseTest.java
```diff
@@ -0,0 +1,57 @@
...
+  @Before
+  public void setUpMocksAndFakes() {
+    MockitoAnnotations.initMocks(this);
+  }
+
+  @Test
+  public void testGetContent() throws IOException {
+    String expectedResponse = "crepecake\nis\ngood!";
+    ByteArrayInputStream responseInputStream =
+        new ByteArrayInputStream(expectedResponse.getBytes());
```
elharo: encoding?

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Connection.java
coollog: Yes, the registry API uses PUT in certain requests. We may need to add a PATCH too but that might not be necessary for now.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Connection.java
coollog: The reason these aren't in static methods right now is because I wanted to provide the closeable interface so that the connection can be auto-closed with the try-with-resource block.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Connection.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /** Sends the request with method PUT. */
+  public Response put(Request request) throws IOException {
+    return send(HttpMethods.PUT, request);
+  }
+
+  /** Sends the request. */
+  private Response send(String httpMethod, Request request) throws IOException {
+    httpResponse =
```
elharo: Instead of implementing Closeable, could we simply close httpResponse here before returning?

coollog: According to the doc of `disconnect`:

```
#disconnect: Close the HTTP response content using {@link #ignore}
#ignore: Closes the content of the HTTP response from {@link #getContent()}, ignoring any content.
```

When disconnect is called, the response is essentially considered dead and nothing can be retrieved from it. Therefore, we can't close the httpResponse before its content is actually read via the returned `Response` object.

So, the use of `Connection` would be like:

```
try (Connection connection = new Connection(url)) {
  Response response = connection.get(request);
  // ... process the response
}
```

crepecake/src/test/java/com/google/cloud/tools/crepecake/http/ConnectionTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.mockito.MockitoAnnotations;
+
+/** Tests for {@link Connection}. */
+public class ConnectionTest {
+
+  @Mock private HttpRequestFactory mockHttpRequestFactory;
+  @Mock private Request mockRequest;
+  @Mock private BlobHttpContent mockBlobHttpContent;
+  @Mock private HttpRequest mockHttpRequest;
+  @Mock private HttpHeaders mockHttpHeaders;
```
elharo: I doubt the headers need to be mocked. Just create an object.

crepecake/src/test/java/com/google/cloud/tools/crepecake/http/ConnectionTest.java
```diff
@@ -0,0 +1,96 @@
...
+        .thenReturn(mockHttpRequest);
+
+    Mockito.when(mockRequest.getBody()).thenReturn(mockBlobHttpContent);
+    Mockito.when(mockRequest.getHeaders()).thenReturn(mockHttpHeaders);
+    Mockito.when(mockHttpRequest.setHeaders(mockHttpHeaders)).thenReturn(mockHttpRequest);
+    Mockito.when(mockHttpRequest.execute()).thenReturn(mockHttpResponse);
+  }
+
+  @Test
+  public void testGet() throws IOException {
```
elharo: These tests really aren't testing very much, and they're very tied to the private internals of the class.

crepecake/src/test/java/com/google/cloud/tools/crepecake/http/ResponseTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.mockito.MockitoAnnotations;
+
+/** Tests for {@link Response}. */
+public class ResponseTest {
+
+  @Mock private HttpResponse httpResponseMock;
+
+  @Before
+  public void setUpMocksAndFakes() {
+    MockitoAnnotations.initMocks(this);
```
elharo: I think there's a RunWith annotation that does this:

crepecake/src/test/java/com/google/cloud/tools/crepecake/http/ConnectionTest.java
coollog: I changed this to be more clear that it's testing that the correct Request is sent with Connection#send.

PR #48, opened 2017-11-27
 Adds image representation.

### LINKED ISSUE #41: Add an image representation.
Consists of 
- layers
- container configuration

coollog:
Fixes #41


crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Image.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Represents an image.
+ *
+ * @param <T> the type of {@link Layer} this image contains
+ */
+public class Image<T extends Layer> {
```
loosebazooka: Maybe I don't understand what an image is, but can't it be different kinds of layers?

coollog: Yea, if the image has different kinds of layers, it would just be Image<Layer>. However, for certain images, like the base image or a built-but-not-pushed image, it might only consist of a certain type of layer. Perhaps templating this not the best way to approach it. I'll remove the template on this for now.

loosebazooka: Well if you're using it to enforce certain paradigms that might be fine.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/ImageLayers.java
```diff
@@ -0,0 +1,53 @@
...
+package com.google.cloud.tools.crepecake.image;
+
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/** Holds the layers for an image. Makes sure that each layer is only added once. */
+class ImageLayers<T extends Layer> {
```
loosebazooka: Does this need to be outside of Image?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Image.java
elharo: I agree with removing the generics.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/ImageLayers.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /** The layers of the image, in the order in which they are applied. */
+  private final List<T> layers = new ArrayList<>();
+
+  /** Keeps track of the layers already added. */
+  private final Set<DescriptorDigest> layerDigests = new HashSet<>();
+
+  /** Returns an immutable copy of the image layers. */
+  List<T> asList() {
```
elharo: ImmutableList in signature

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Image.java
```diff
@@ -0,0 +1,74 @@
...
+
+  public Map<String, String> getEnvironmentMap() {
+    return ImmutableMap.copyOf(environmentMap);
+  }
+
+  public void setEnvironmentVariable(String name, String value) {
+    environmentMap.put(name, value);
+  }
+
+  public List<String> getEntrypoint() {
```
elharo: ImmutableList<String> or return a mutable copy. It's a design flaw in the Collections API that an ImmutableList cannot stand in for its supertype because it doesn't fulfill the contract of a List. :-(

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Image.java
```diff
@@ -0,0 +1,74 @@
...
+  public Image() {
+    layers = new ImageLayers<>();
+  }
+
+  @VisibleForTesting
+  Image(ImageLayers<T> imageLayers) {
+    layers = imageLayers;
+  }
+
+  public Map<String, String> getEnvironmentMap() {
```
elharo: ImmutableMap

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/ImageLayersTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+/** Tests for {@link ImageLayers}. */
+public class ImageLayersTest {
+
+  @Mock private DescriptorDigest mockDescriptorDigest1;
```
elharo: This is a lot of mocks. Could any of these be replaced by actual objects with known values?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Image.java
```diff
@@ -0,0 +1,74 @@
...
+
+  /** Initial command to run when running the image. */
+  private List<String> entrypoint;
+
+  public Image() {
+    layers = new ImageLayers<>();
+  }
+
+  @VisibleForTesting
+  Image(ImageLayers<T> imageLayers) {
```
elharo: You shouldn't need this constrcutor. At most you need a getter method on the field, and probably not even that.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/ImageLayers.java
coollog: This will be used in other places such as the cache management classes for just holding a list of layers.

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/ImageLayersTest.java
coollog: Is having a lot of mocks bad practice? I needed many different blob descriptors because ImageLayers prevents the duplicate layer digests.

elharo: Not in and of itself, but the more you mock, the less you test. In general test data that can be an actual object, should be an actual object. I'm not sure if that applies here or not.

Also, not all mocks have to be fields. Mocks referenced in only one method should be local variables. Here it looks like the descriptor digests can all move into `setUpFakes`.

PR #49, opened 2017-11-29
 Adds LayerBuilder to build a layer from a set of files.

coollog:


crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/LayerBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.cloud.tools.crepecake.tar.TarStreamBuilder;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+
+/** Builds an {@link UnwrittenLayer} from files. */
+public class LayerBuilder {
```
loosebazooka: This seems like a little light of a wrapper around TarStreamBuilder? There's a lot of late evaluation going on, so I might have this wrong. Is there potential to mimic the `Blobs` model for `Layers`? `Layers.from(TarStreamBuilder xyz)?` or are trying to hide `TarStreamBuilder`?

crepecake/src/main/java/com/google/cloud/tools/crepecake/tar/GzipCompressor.java
```diff
@@ -0,0 +1,17 @@
+package com.google.cloud.tools.crepecake.tar;
+
+import java.io.OutputStream;
+import org.apache.commons.compress.compressors.CompressorException;
+import org.apache.commons.compress.compressors.CompressorStreamFactory;
+
+/** Wraps {@link OutputStream}s with GZIP compression. */
+public abstract class GzipCompressor {
```
elharo: Could you just use GZipOutputStream from the JDK?

crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/LayerBuilder.java
coollog: Yeah, this is just trying to hide `TarStreamBuilder` since building a layer should be agnostic to how the archive is built. Rather than having to add files to a layer by constructing a `TarArchiveEntry`, the layer should be constructed with the file to add and the path on the final image filesystem the file belongs to.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/LayerBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+   * @param file the file to add
+   * @param path the path of the file in the partial filesystem changeset
+   */
+  public void addFile(File file, String path) {
+    filesystemEntries.add(new TarArchiveEntry(file, path));
+  }
+
+  /** Builds and returns the layer. */
+  public UnwrittenLayer build() {
+    TarStreamBuilder tarStreamBuilder = tarStreamBuilderSupplier.get();
```
elharo: why not just new TarStreamBuilder()?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/LayerBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+    // Adds all the files to a tar.gzip stream.
+    for (TarArchiveEntry entry : filesystemEntries) {
+      tarStreamBuilder.addEntry(entry);
+    }
+
+    return new UnwrittenLayer(tarStreamBuilder.toBlob());
+  }
+
+  @VisibleForTesting
+  LayerBuilder(Supplier<TarStreamBuilder> tarStreamBuilderSupplier) {
```
elharo: It shouldn't be necessary to contort the API like this for testing.

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/LayerBuilderTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    // Fake files to build into the layer.
+    List<TarArchiveEntry> fileEntries =
+        Arrays.asList(
+            new TarArchiveEntry(new File("fileA"), "/path/to/fileA"),
+            new TarArchiveEntry(new File("directory/fileB"), "/path/to/directory/fileB"),
+            new TarArchiveEntry(new File("directory/fileC"), "/path/to/directory/fileC"),
+            new TarArchiveEntry(new File("directory/"), "/path/to/directory/"));
+
+    LayerBuilder layerBuilder = new LayerBuilder(() -> mockTarStreamBuilder);
```
elharo: I think if you democked this the test would be more useful. We can certainly write a real tar file during a test.

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/LayerBuilderTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+/** Tests for {@link LayerBuilder}. */
+public class LayerBuilderTest {
+
+  @Mock private Blob mockBlob;
```
elharo: Not sure you need to mock the Blob. We should be able to use a file blob or an empty blob here.

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/LayerBuilderTest.java
```diff
@@ -0,0 +1,71 @@
...
+  @Mock private TarStreamBuilder mockTarStreamBuilder;
+
+  @Before
+  public void setUpMocks() throws IOException {
+    MockitoAnnotations.initMocks(this);
```
elharo: If you need mocks at all (and I'm not sure you do here) then use the @RunWith(MockitoJUnitRunner) annotation

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/LayerBuilderTest.java
coollog: Since the TarStreamBuilderTest already verifies that the TarStreamBuilder builds the right tar, this test was intended to just test that the file entries were added to the TarStreamBuilder correctly. I'll remove all of the Supplier and API-changing stuff.

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/LayerBuilderTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+/** Tests for {@link LayerBuilder}. */
+@RunWith(MockitoJUnitRunner.class)
+public class LayerBuilderTest {
+
+  @Mock private TarStreamBuilder mockTarStreamBuilder;
+
+  @InjectMocks private LayerBuilder layerBuilder;
```
elharo: Interesting. I haven't seen this technique before. 

That said, I'm still not sure you have to mock this. Could we use real files instead of fake files and then verify the real tar that's built? That would be a more comprehensive test.

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/LayerBuilderTest.java
```diff
@@ -0,0 +1,106 @@
...
+    Path layerDirectory = Paths.get(Resources.getResource("layer").toURI());
+
+    LayerBuilder layerBuilder = new LayerBuilder();
+
+    // Adds each file in the layer directory to the layer builder.
+    Files.walk(layerDirectory)
+        .filter(path -> !path.equals(layerDirectory))
+        .forEach(
+            path ->
+                layerBuilder.addFile(path.toFile(), layerDirectory.relativize(path).toString()));
```
elharo: This is a bit strange. Why do we need to pass the file and the string path to addFile. They seem to contain the same information.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/LayerBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+  /** The partial filesystem changeset to build the layer with. */
+  private final List<TarArchiveEntry> filesystemEntries = new ArrayList<>();
+
+  /**
+   * Prepares a file to be built into the layer.
+   *
+   * @param file the file to add
+   * @param path the path of the file in the partial filesystem changeset
+   */
+  public void addFile(File file, String path) {
```
elharo: Noticed this in the test. Are these really two independent arguments? Or is the path always formed from the file in an obvious way?

crepecake/src/main/java/com/google/cloud/tools/crepecake/tar/TarStreamBuilder.java
```diff
@@ -44,22 +43,17 @@ private static void writeEntriesAsTarArchive(
         new TarArchiveOutputStream(tarByteStream)) {
       for (TarArchiveEntry entry : entries) {
         tarArchiveOutputStream.putArchiveEntry(entry);
-        InputStream contentStream = new BufferedInputStream(new FileInputStream(entry.getFile()));
-        ByteStreams.copy(contentStream, tarArchiveOutputStream);
+        if (entry.isFile()) {
```
loosebazooka: Are directories just created as needed when adding files?

coollog: Here, directories are still added as `putArchiveEntry` puts the directory into the archive stream. Directories don't have content so that's why this code is only for regular files.

PR #50, opened 2017-12-04
Changes ByteHasher to ByteHashBuilder that can build a hash over multiple calls.

coollog:
This is to better fit a design of streaming blobs where the blob hash is not computed until the full blob has been streamed out.

### DIFF
--- a/crepecake/src/main/java/com/google/cloud/tools/crepecake/hash/ByteHashBuilder.java
+++ b/crepecake/src/main/java/com/google/cloud/tools/crepecake/hash/ByteHashBuilder.java
@@ -66,6 +66,7 @@ public class ByteHashBuilder extends OutputStream {
     } catch (NoSuchAlgorithmException ex) {
       throw new IOException(ex);
     }
+    totalBytes += length;
   }
 
   @Override
@@ -75,6 +76,7 @@ public class ByteHashBuilder extends OutputStream {
     } catch (NoSuchAlgorithmException ex) {
       throw new IOException(ex);
     }
+    totalBytes += data.length;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


coollog:
@elharo 
Changing this to use java.security.DigestOutputStream as recommended

crepecake/src/main/java/com/google/cloud/tools/crepecake/hash/ByteHashBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+    byte[] singleByteArray = {(byte) (singleByte & 0xff)};
+    try {
+      append(singleByteArray, 0, 1);
+    } catch (NoSuchAlgorithmException ex) {
+      throw new IOException(ex);
+    }
+  }
+
+  /** Appends data to the bytes the hash. */
+  private void append(byte[] data, int offset, int length) throws NoSuchAlgorithmException {
```
elharo: I don't think this throws NoSuchAlgorithmException. That seems to only be in the constructor.

crepecake/src/main/java/com/google/cloud/tools/crepecake/hash/ByteHashBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Generates SHA-256 hashes in hexadecimal. Can act as an {@link OutputStream} that captures the
+ * bytes to hash.
+ */
+public class ByteHashBuilder extends OutputStream {
+
+  private static final String SHA_256_ALGORITHM = "SHA-256";
+
+  private final MessageDigest messageDigest = MessageDigest.getInstance(SHA_256_ALGORITHM);
```
elharo: This throws NoSuchAlgorithmException

coollog:
Closing this to have the changes be in #42 since it's used directly there.

PR #53, opened 2017-12-12
Removes DigestException from BlobStream#writeTo.

coollog:
Realized it doesn't make sense for writeTo to throw a DigestException since the hashing should work fine if the algorithm is implemented correctly.


coollog:
@elharo @loosebazooka This can be merged in before any of the rest.

loosebazooka:
Do you mean that it would be our fault if the hashing isn't implemented correctly? Should it be an assertion error?

coollog:
@loosebazooka It means it wouldn't be our fault if the hashing isn't implemented correctly. `DescriptorDigest` is where the `DigestException` is thrown, and the intent of that was to catch any invalid digests received over HTTP or from the user side. Digests generated by the JDK should always be valid.

coollog:
Closing since these changes were added in #43

Issue #52, opened 2017-12-10
Add cacher.


coollog:
Caches all layers (content and diffId) and last modified times for the application layers.

Cache the base image manifest as well to know to not repull the base image.

- [x] Metadata #56 
- [x] Metadata JSON template #70 
- [x] Translator for metadata JSON #71 
- [x] Initialize cache with root directory #71 
- [x] Cache writer #79
- [x] Cache checker #81 
- [x] Cacher reader #82
- [x] Cache pruner? (TODO for https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/111)

PR #56, opened 2017-12-12
Adds cache management classes for holding cache information.

coollog:
Part of #52


crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheHelper.java
```diff
@@ -0,0 +1,48 @@
...
+package com.google.cloud.tools.crepecake.cache;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.io.File;
+import java.nio.file.Path;
+
+/** Methods for getting static cache properties. */
+class CacheHelper {
+
+  @VisibleForTesting static final String LAYER_FILE_EXTENSION = ".tar.gz";
```
elharo: There's no need to make these visible for testing. Just use the string constants in the test, which also helps test the values.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheHelper.java
```diff
@@ -0,0 +1,48 @@
...
+  @VisibleForTesting static final String RESOURCES_LAYER_NAME = "resources";
+  @VisibleForTesting static final String CLASSES_LAYER_NAME = "classes";
+
+  static File getLayerFile(Path cacheDirectory, String layerName) {
+    return cacheDirectory.resolve(layerName + LAYER_FILE_EXTENSION).toFile();
+  }
+
+  static String getNameForApplicationLayer(ApplicationLayerType layerType) {
+    switch (layerType) {
+      case DEPENDENCIES:
```
elharo: or you could give the ApplicationLayerType enum a getName method. That would be a little simpler and more OO.

crepecake/src/test/java/com/google/cloud/tools/crepecake/cache/TimestampedCachedLayerTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.mockito.MockitoAnnotations;
+
+/** Tests for {@link TimestampedCachedLayer}. */
+public class TimestampedCachedLayerTest {
+
+  @Mock private CachedLayer mockCachedLayer;
+
+  @Before
+  public void setUpMocks() {
+    MockitoAnnotations.initMocks(this);
```
elharo: use the MockitoJUnitRunner

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/UnwrittenLayerTest.java
```diff
@@ -52,6 +53,7 @@ public void setUpMocks() throws IOException {
     Mockito.when(mockBlobDescriptor.getDigest()).thenReturn(mockDiffId);
   }
 
+  // TODO: Test writeCompressedBlobTo
```
elharo: do this or file it as a a bug

coollog: This is no longer relevant since UnwrittenLayer only has an uncompressed blob. Removed

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheHelper.java
coollog: These constants are actually the file names for the layers, which should probably live in some configuration class with any other cache configuration. Perhaps this CacheHelper class should be named CacheConfiguration.

elharo: something like 

```
return layerType.getName()
```

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheHelper.java
```diff
@@ -0,0 +1,32 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.cache;
+
+import java.io.File;
+import java.nio.file.Path;
+
+/** Methods for getting static cache properties. */
+class CacheHelper {
```
loosebazooka: In general, I think we should try not to use names like `*Helper`, maybe we're not categorizing it's purpose correctly. Is there something we can call this that makes more sense? CacheFiles? or something more specific?

loosebazooka:
Is the cache layer location configurable? Can we potentially share a cache with a docker daemon? Or at least share a cache between all a user's projects on a single machine?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
```diff
@@ -0,0 +1,52 @@
...
+package com.google.cloud.tools.crepecake.cache;
+
+import com.google.cloud.tools.crepecake.image.DuplicateLayerException;
+import com.google.cloud.tools.crepecake.image.ImageLayers;
+import com.google.cloud.tools.crepecake.image.LayerPropertyNotFoundException;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/** Hold information for what data the cache holds. */
```
loosebazooka: This comment is a little weirdly worded?

Also, how is CacheMetadata going to be used? Is this some reflection of the filesystem (and the fs is the source of truth?) or is it going to be edited to written back down?

coollog: The cache will store all the cached blobs as well as a `metadata.json` file that has information describing each blob in the cache. The `CacheMetadata` class will be loaded from `metadata.json` and then changed over the course of the build and written back to the fs.

coollog:
Yea, the cache location will be configurable. From looking around, it seems inpractical to share the cache with a Docker daemon since at least on macOS, the Docker daemon stores all its images in one large file at `~/Library/Containers/com.docker.docker/Data/com.docker.driver.amd64-linux/Docker.qcow2`. However, it is feasible to share a cache between all a user's projects on a single machine.

PR #67, opened 2017-12-14
Adds authenticators for Docker registry.

### LINKED ISSUE #57: Add pull image authorization credential manager

coollog:
Fixes #57 and #58


crepecake/build.gradle
```diff
@@ -13,9 +13,25 @@ repositories {
...
+
+configurations {
+    integrationTestCompile.extendsFrom testCompile
+    integrationTestRuntime.extendsFrom testRuntime
+}
+
 dependencies {
     compile gradleApi()
-    compile 'com.google.http-client:google-http-client:1.23.0'
+    compile 'com.google.http-client:google-http-client-jackson2:1.23.0'
```
elharo: we don't need http-client any more?

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/DockerCredentialRetrieverIntegrationTest.java
```diff
@@ -0,0 +1,39 @@
+package com.google.cloud.tools.crepecake.registry;
```
elharo: copyright

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/DockerCredentialRetrieverIntegrationTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+/** Integration tests for {@link DockerCredentialRetriever}. */
+public class DockerCredentialRetrieverIntegrationTest {
+
+  @Test
+  public void testRetrieveGCR() throws IOException {
+    DockerCredentialRetriever dockerCredentialRetriever =
+        new DockerCredentialRetriever("gcr.io", "gcloud");
+
+    Authorization authorization;
```
elharo: move all the non-error code inside the try

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/RegistryAuthenticatorIntegrationTest.java
```diff
@@ -0,0 +1,19 @@
+package com.google.cloud.tools.crepecake.registry;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Authorization.java
```diff
@@ -0,0 +1,27 @@
+package com.google.cloud.tools.crepecake.http;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Authorization.java
```diff
@@ -0,0 +1,27 @@
...
+  private final String scheme;
+  private final String token;
+
+  Authorization(String scheme, String token) {
+    this.scheme = scheme;
+    this.token = token;
+  }
+
+  /** @return the {@code Authorization} header as a two-element list: [scheme, token] */
+  public List<String> asList() {
```
elharo: why do you need this?

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Authorizations.java
```diff
@@ -0,0 +1,13 @@
+package com.google.cloud.tools.crepecake.http;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Request.java
```diff
@@ -38,6 +38,12 @@ public BlobHttpContent getHttpContent() {
     return body;
   }
 
+  // TODO: Add to ConnectionTest.
+  public Request setAuthorization(Authorization authorization) {
+    headers.setAuthorization(authorization.asList());
```
elharo: The HttpHeaders setAuthorization method isn't very well documented, I'm afraid. Hard to tell what's happening.

I do suggest converting to a list here rather than in Authorization.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/DockerCredentialRetriever.java
```diff
@@ -0,0 +1,62 @@
+package com.google.cloud.tools.crepecake.registry;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryAuthenticationFailedException.java
```diff
@@ -0,0 +1,9 @@
+package com.google.cloud.tools.crepecake.registry;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryAuthenticator.java
```diff
@@ -0,0 +1,61 @@
+package com.google.cloud.tools.crepecake.registry;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryAuthenticators.java
```diff
@@ -0,0 +1,17 @@
+package com.google.cloud.tools.crepecake.registry;
```
elharo: copyright

crepecake/build.gradle
coollog: It seems that -jackson2 comes with the core `google-http-client` artifact.

elharo: I'm not sure how gradle works here. In google3/blaze we try to be explicit about dependencies. That is, if we directly reference classes from library A and library B, we include both libraries A and B in our BUILD file, even if B pulls in A transitively. However if we reference B which depends on A, but do not explicitly reference classes from A, then we don't include A in our dependency list. This makes the build more consistent as dependencies evolve. However perhaps gradle does things differently.

coollog: I'll add it back it then. @loosebazooka for clarification on what the best practice here is for Gradle

loosebazooka: I'd agree with Elliotte here

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/DockerCredentialRetrieverIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+/** Integration tests for {@link DockerCredentialRetriever}. */
+public class DockerCredentialRetrieverIntegrationTest {
+
+  /**
+   * Tests retrieval via {@code docker-credential-gcloud} CLI.
+   *
+   * <p>This test will ignore errors if
+   *
+   * <ul>
```
loosebazooka: I think we should fail if the test environment is not set up correctly, no?

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/Blobs.java
```diff
@@ -20,7 +20,7 @@
 import java.io.InputStream;
 
 /** Static initializers for {@link Blob}. */
-public class Blobs {
+public abstract class Blobs {
```
loosebazooka: Do we expect people to extend this? Why not a private constructor to prevent instantiation?

coollog:
Removed integration test invocation in Kokoro script. Will need to add running integration tests on commits to master.

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/DockerCredentialRetrieverIntegrationTest.java
```diff
@@ -0,0 +1,53 @@
...
+      DockerCredentialRetriever dockerCredentialRetriever =
+          new DockerCredentialRetriever("gcr.io", "gcr");
+
+      Authorization authorization = dockerCredentialRetriever.retrieve();
+
+      // Checks that some token was received.
+      Assert.assertTrue(0 < authorization.getToken().length());
+
+    } catch (MismatchedInputException ex) {
+      if (!ex.getMessage().contains("No content to map due to end-of-input")) {
```
elharo: why do you allow these failures?

coollog: Here, I am giving a more descriptive error for test failure cases where it's not necessarily the code is invalid but rather the environment that does not have the necessary components to run the test correctly.

elharo: That's an antipattern. It does come up occasionally but please try really, really hard not to do this. If you have to, then the test should be reported as skipped, not failed. Assume class can help with this. 

But probably we should just fail if the environment is deficient.

coollog: Do you mean it's an antipattern to fail when the environment is deficient or to catch exceptions just for providing a different exception message?

elharo: Failing when the environment is deficient. In general, tests should not be environment dependent. 

On occasion we have tests that only apply in one platform (e.g. Windows) but in that case if you run them on Unix they should be marked skipped, not passed.

Other conditions like bad network access should be failures.

coollog: @loosebazooka I'll revert that change and mark as skipped then?

elharo: In this case I don't think you've justified marking it as skipped. This looks like a straight test failure to me.

coollog: So here, if the credential store doesn't have credentials for a certain server URL, that means that the environment wasn't set up correctly (ie the credentials for that server URL wasn't put into the credential store).

crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonHelper.java
```diff
@@ -44,9 +44,9 @@
  *
  * @see <a href="https://github.com/FasterXML/jackson">https://github.com/FasterXML/jackson</a>
  */
-public class JsonHelper {
+public abstract class JsonHelper {
```
loosebazooka: I think guidlines for non-instantiable classes is to use private constructor.

crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonHelper.java
```diff
@@ -46,7 +46,7 @@
  */
 public class JsonHelper {
```
loosebazooka: If this class is just dealing with JsonTemplateSerialization, I think we should indicate that in the name. Helper is too generic.

crepecake/src/main/java/com/google/cloud/tools/crepecake/json/JsonHelper.java
```diff
@@ -46,7 +46,7 @@
...
 
-  private static final ObjectMapper objectMapper = new ObjectMapper();
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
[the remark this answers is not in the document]
coollog: The class itself is not mutable but we want to use it in an immutable fashion here.

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/DockerCredentialRetrieverIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+      Authorization authorization = dockerCredentialRetriever.retrieve();
+
+      // Checks that some token was received.
+      Assert.assertTrue(0 < authorization.getToken().length());
+
+    } catch (NonexistentServerUrlDockerCredentialHelperException
+        | NonexistentDockerCredentialHelperException ex) {
+
+      Assume.assumeNoException("Skipping because docker-credential-gcr CLI not set up", ex);
```
elharo: Again, I really think this should be a straight failure. If this happens, the build should fail and the dev should set up the credential CLI. We shouldn't be running this test without that.

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/DockerCredentialRetrieverIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+    try {
+      DockerCredentialRetriever fakeDockerCredentialRetriever =
+          new DockerCredentialRetriever("", "fake-cloud-provider");
+
+      fakeDockerCredentialRetriever.retrieve();
+
+      Assert.fail("Retrieve should have failed for nonexistent credential helper");
+
+    } catch (NonexistentDockerCredentialHelperException ex) {
+      Assert.assertEquals(
```
elharo: just fail

coollog: The intention of this test is to check that this exception gets thrown for nonexistent credential helpers.

crepecake/build.gradle
```diff
@@ -27,13 +44,24 @@ task wrapper(type: Wrapper) {
     gradleVersion = "4.3.1"
 }
 
+// Integration tests must be run explicitly
```
loosebazooka: so we're not running them on CI at the moment?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/NonexistentServerUrlDockerCredentialHelperException.java
```diff
@@ -0,0 +1,25 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.registry;
+
+/** Thrown because the credential helper does not have credentials for the specified server URL. */
+public class NonexistentServerUrlDockerCredentialHelperException extends Exception {
+
+  NonexistentServerUrlDockerCredentialHelperException(String serverUrl) {
+    super("The credential store has nothing for server URL of " + serverUrl);
```
loosebazooka: If this message is going to be visible to users, we should be more explicit.
Is something like this accurate : `The credential helper (name) has no credential for server URL: serverUrl`?

Issue #59, opened 2017-12-13
Add registry API client.


coollog:
- [x] image manifest puller #68 
- [x] blob puller https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/pull/85
- [x] image manifest pusher #88 
- [x] blob pusher #93
- [x] blob existence checker #108 
- [x] authentication method retriever #95

PR #68, opened 2017-12-18
 Adds image manifest puller.

coollog:
Step 1 of #59

This one's a bit big. Should have probably split this into multiple PRs.


crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/ManifestPullerIntegrationTest.java
```diff
@@ -0,0 +1,122 @@
...
+      throws IOException, RegistryErrorException, RegistryUnauthorizedException,
+          RegistryTooManyRequestsException, UnknownManifestFormatException {
+    if (!localRegistryRunning) {
+      return;
+    }
+
+    ManifestPuller manifestPuller = new ManifestPuller(null, "localhost:5000", "busybox");
+    ManifestTemplateHolder manifestTemplateHolder = manifestPuller.pull("latest");
+
+    if (manifestTemplateHolder.isV21()) {
```
elharo: Conditional logic in unit tests is suspect. One or the other of these should be true. Just test that.

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/ManifestPullerIntegrationTest.java
```diff
@@ -0,0 +1,122 @@
...
+    String removeRegistryContainerCommand = "docker rm -v registry";
+    Runtime.getRuntime().exec(removeRegistryContainerCommand).waitFor();
+  }
+
+  @Test
+  public void testPull()
+      throws IOException, RegistryErrorException, RegistryUnauthorizedException,
+          RegistryTooManyRequestsException, UnknownManifestFormatException {
+    if (!localRegistryRunning) {
+      return;
```
elharo: No, this is a test failure. It's dangerous to skip tests like this.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/HttpStatusCodes.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.http;
+
+/** Static constants for HTTP status codes. */
+public abstract class HttpStatusCodes {
```
elharo: These constants exist in HttpURLConnection and several other classes. No need to include it here.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/json/ManifestTemplateHolder.java
```diff
@@ -0,0 +1,80 @@
...
+    this.v22ManifestTemplate = null;
+  }
+
+  private ManifestTemplateHolder(@Nonnull V22ManifestTemplate v22ManifestTemplate) {
+    this.v21ManifestTemplate = null;
+    this.v22ManifestTemplate = v22ManifestTemplate;
+  }
+
+  public boolean isV21() {
+    return null != v21ManifestTemplate;
```
elharo: flip; null == x is a C convention that doesn't apply in Java

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/json/ManifestTemplateHolder.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  public boolean isV22() {
+    return null != v22ManifestTemplate;
```
elharo: flip

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/json/ManifestTemplateHolder.java
```diff
@@ -0,0 +1,80 @@
...
+
+package com.google.cloud.tools.crepecake.image.json;
+
+import com.google.cloud.tools.crepecake.json.JsonTemplateMapper;
+import com.google.common.base.Preconditions;
+import java.io.IOException;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/** Holds one of {@link V21ManifestTemplate} or {@link V22ManifestTemplate}. */
```
elharo: Might make sense to have subclasses for each version instead.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/ManifestPuller.java
```diff
@@ -0,0 +1,99 @@
...
+        request.setAuthorization(authorization);
+      }
+      Response response = connection.get(request);
+      String responseString = response.getBody().writeToString();
+
+      return ManifestTemplateHolder.fromJson(responseString);
+
+    } catch (HttpResponseException ex) {
+      switch (ex.getStatusCode()) {
+        case HttpStatusCodes.BAD_REQUEST:
```
elharo: switch fallthrough is very error prone

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/HttpStatusCodes.java
loosebazooka: private constructor.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/Blob.java
```diff
@@ -29,4 +31,11 @@
    * @return the {@link BlobDescriptor} of the written BLOB
    */
   BlobDescriptor writeTo(OutputStream outputStream) throws IOException;
+
+  /** Writes the BLOB to a string. */
+  default String writeToString() throws IOException {
```
loosebazooka: I don't think we want to use default here. While it's nice, I think it throws us back into the abstract class pattern. Can this just part of Blobs?

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobDescriptor.java
```diff
@@ -67,18 +78,18 @@ public long getSize() {
...
-   *   <li>{@code digest}s are not null and equal
-   *   <li>{@code size} is equal
+   *   <li>{@code digest}s are not null and equal, and
+   *   <li>{@code size}s are non-negative and equal
    * </ol>
    */
   @Override
   public boolean equals(Object obj) {
-    if (digest == null || !(obj instanceof BlobDescriptor)) {
+    if (digest == null || size < 0 || !(obj instanceof BlobDescriptor)) {
```
elharo: Given the weird logic here when size or digest is missing, I think you need to first check for identity. I.e.

if (obj == this) return true;

Otherwise this violates the contract of equals.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/Blobs.java
```diff
@@ -16,10 +16,13 @@
 
 package com.google.cloud.tools.crepecake.blob;
 
+import com.google.common.base.Charsets;
```
elharo: StandardCharsets is preferred now.
Guava Charsets is a legacy of pre-Java 7.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/ManifestPuller.java
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Instantiates a {@link ManifestTemplate} from a JSON string. This checks the {@code
+   * schemaVersion} field of the JSON to determine which manifest version to use.
+   */
+  public static ManifestTemplate getManifestTemplateFromJson(String jsonString)
+      throws IOException, UnknownManifestFormatException {
+    ManifestTemplate manifestTemplate =
+        JsonTemplateMapper.readJson(jsonString, ManifestTemplate.class);
+
+    switch (manifestTemplate.getSchemaVersion()) {
```
elharo: could this be pushed into the ManifestTemplate class to avoid the switch statement?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/ManifestPuller.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  public ManifestTemplate pull(String imageTag)
+      throws IOException, RegistryErrorException, RegistryUnauthorizedException,
+          UnknownManifestFormatException {
+    URL pullUrl = getApiRoute("/manifests/" + imageTag);
+
+    try (Connection connection = new Connection(pullUrl)) {
+      Request request = new Request();
+      if (null != authorization) {
```
elharo: flip

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/Blobs.java
coollog: Filed #74 to change this in other parts as well.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/ManifestPuller.java
coollog: `V21ManifestTemplate` and `V22ManifestTemplate` are subclasses of `ManifestTemplate`. Perhaps I should make another class called `ManifestTemplates` with a `#fromJson` method that does this?

elharo: What you're doing here is close to 


switch (type) {
  case type1:
   doThing1
  case Type2:
    doThing2

This pattern is usually a failure to apply OO. The OO equivalent is

type.doThing()

In this case what seems called for is

manifestTemplate.readJson()

If the `manifestTemplate` variable is a V21ManifestTemplate then it will use that. If it's V22ManifestTemplate then it uses that.

coollog: I'm probably going about this the wrong way but it seems that in this situation, polymorphism doesn't apply since it is first parsing the JSON as a `ManifestTemplate` object and then re-parsing the JSON as one of the subclasses according to a field in the `ManifestTemplate` (not resolved to one of the subclasses yet).

elharo: That is quite weird and unconventional. It may suggest that data binding is not the best way to read these files.

coollog: Should I do something like described in https://stackoverflow.com/questions/26190851/get-single-field-from-json-using-jackson?

elharo: That's one way of doing it, I guess.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobDescriptor.java
```diff
@@ -46,14 +47,24 @@ public BlobDescriptor(long size, DescriptorDigest digest) {
     this.digest = digest;
   }
 
+  /** Initialize with just size. */
```
loosebazooka: when would one partially initialize a blob?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/json/ManifestTemplate.java
```diff
@@ -0,0 +1,33 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.image.json;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.google.cloud.tools.crepecake.json.JsonTemplate;
+
+/**
+ * JSON template for just the {@code schemaVersion} field of image manifests. A {@code schemaVersion} of {@code 1} indicates that the JSON should be parsed as {@link V21ManifestTemplate}, whereas a {@code schemaVersion] of {@code 2} indicates that the JSON should be parsed as {@link V22ManifestTemplate}.
```
loosebazooka: Is this comment line formatted correctly? Seems a little long?

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobDescriptor.java
coollog: This is a BlobDescriptor that is used in FileBlob and InputStreamBlob, where the bytes are counted while piping but no hash is calculated. It is also used in EmptyBlob and especially StringBlob, where the digest doesn't apply, like an image manifest.

loosebazooka: It's hard for me to tell, but do you think this is an indication that we could have overgeneralized our classes?

coollog: Yeah, that's definitely possible. I'll file an issue to think more about this. BlobDescriptor was intended to represent the content descriptors in an image manifest, which have size and digest. It was put into one class to return either or both from Blob#writeTo so that the digest and size of the blob can be captured upon write.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/ManifestPuller.java
```diff
@@ -0,0 +1,128 @@
...
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+import javax.annotation.Nullable;
+
+/** Registry interface for pulling an image's manifest. */
+public class ManifestPuller {
+
+  // TODO: This should be configurable.
+  private static final String PROTOCOL = "http";
```
loosebazooka: should we be pulling over https?

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Request.java
```diff
@@ -24,17 +24,17 @@
 public class Request {
```
loosebazooka: kinda looks like Request should have a builder and be immutable?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryErrorExceptionBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+    String reason = getReason(errorEntry.getCode(), errorEntry.getMessage());
+    if (!firstErrorReason) {
+      errorMessageBuilder.append(", ");
+    }
+    errorMessageBuilder.append(reason);
+    firstErrorReason = false;
+    return this;
+  }
+
+  RegistryErrorException toRegistryHttpException() {
```
loosebazooka: I think it's clear what you're trying to build here. Could this just be "build"?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/ManifestPuller.java
coollog: As discussed, https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/78 filed to address this.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Request.java
coollog: Changed to builder pattern.

coollog:
Yep, the unfixed comments have been separately filed as issues.

elharo:
@coollog Can this be merged?

coollog:
@elharo Yes, merging now.

PR #70, opened 2017-12-20
Adds cache metadata JSON template.

coollog:
Part of #52

### DIFF
--- a/crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
+++ b/crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
@@ -91,11 +91,11 @@ class CacheMetadataLayerObjectTemplate extends JsonTemplate {
   }
 
   public List<String> getSourceDirectories() throws CacheMetadataCorruptedException {
-    return getPropertiesNonnull().sourceDirectories;
+    return getPropertiesNonNull().sourceDirectories;
   }
 
   public long getLastModifiedTime() throws CacheMetadataCorruptedException {
-    return getPropertiesNonnull().lastModifiedTime;
+    return getPropertiesNonNull().lastModifiedTime;
   }
 
   public CacheMetadataLayerObjectTemplate setType(CachedLayerType type) {
@@ -125,20 +125,24 @@ class CacheMetadataLayerObjectTemplate extends JsonTemplate {
 
   public CacheMetadataLayerObjectTemplate setSourceDirectories(List<String> sourceDirectories) {
     checkTypeValidForProperties();
-    createPropertiesIfNull().sourceDirectories = sourceDirectories;
+    createPropertiesIfNull();
+    assert properties != null;
+    properties.sourceDirectories = sourceDirectories;
     return this;
   }
 
   public CacheMetadataLayerObjectTemplate setLastModifiedTime(long lastModifiedTime) {
     checkTypeValidForProperties();
-    createPropertiesIfNull().lastModifiedTime = lastModifiedTime;
+    createPropertiesIfNull();
+    assert properties != null;
+    properties.lastModifiedTime = lastModifiedTime;
     return this;
   }
 
   /** Returns properties if non-null; otherwise throws the appropriate exception. */
-  private PropertiesObject getPropertiesNonnull() throws CacheMetadataCorruptedException {
+  private PropertiesObject getPropertiesNonNull() throws CacheMetadataCorruptedException {
     checkTypeValidForProperties();
-    if (null == properties) {
+    if (properties == null) {
       throw new CacheMetadataCorruptedException("Properties not found for application layer type");
     }
     return properties;
@@ -155,10 +159,9 @@ class CacheMetadataLayerObjectTemplate extends JsonTemplate {
   }
 
   /** Instantiates {@link #properties} if it's {@code null}. */
-  private PropertiesObject createPropertiesIfNull() {
-    if (null != properties) {
-      return properties;
+  private void createPropertiesIfNull() {
+    if (properties == null) {
+      properties = new PropertiesObject();
     }
-    return properties = new PropertiesObject();
   }
 }
--- a/crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataTemplate.java
+++ b/crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataTemplate.java
@@ -47,7 +47,7 @@ import java.util.List;
  *       },
  *       "existsOn": ["gcr.io/my-project/my-image"],
  *       "properties": {
- *         "sourceDirectory": "build/classes",
+ *         "sourceDirectories": ["build/classes"],
  *         "lastModifiedTime": 255073580723571
  *       }
  *     },


crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
```diff
@@ -31,7 +31,7 @@
 
   private final ImageLayers<TimestampedCachedLayer> baseImageLayers = new ImageLayers<>();
 
-  private Map<ApplicationLayerType, TimestampedCachedLayer> applicationLayers = new HashMap<>();
+  private Map<CachedLayerType, TimestampedCachedLayer> applicationLayers = new HashMap<>();
```
elharo: There can be only one layer per type?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
```diff
@@ -0,0 +1,161 @@
...
+        && type != CachedLayerType.RESOURCES
+        && type != CachedLayerType.CLASSES) {
+      throw new IllegalStateException(
+          "Properties is not a valid field for non-application layer type");
+    }
+  }
+
+  /** Instantiates {@link #properties} if it's {@code null}. */
+  private PropertiesObject createPropertiesIfNull() {
+    return properties = null == properties ? new PropertiesObject() : properties;
```
elharo: Too convoluted to follow. Please break up into multiple statements.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
coollog: This class will be changed to match the new format. I added a TODO note. The change will be added in a followup PR that includes the translator to/from the cache metadata and its JSON.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
```diff
@@ -0,0 +1,164 @@
...
+  public CacheMetadataLayerObjectTemplate setLastModifiedTime(long lastModifiedTime) {
+    checkTypeValidForProperties();
+    createPropertiesIfNull().lastModifiedTime = lastModifiedTime;
+    return this;
+  }
+
+  /** Returns properties if non-null; otherwise throws the appropriate exception. */
+  private PropertiesObject getPropertiesNonnull() throws CacheMetadataCorruptedException {
+    checkTypeValidForProperties();
+    if (null == properties) {
```
elharo: no need to do this in Java since you can't use an assignment operator here
properties == null
is more conventional

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  public CacheMetadataLayerObjectTemplate setLastModifiedTime(long lastModifiedTime) {
+    checkTypeValidForProperties();
+    createPropertiesIfNull().lastModifiedTime = lastModifiedTime;
+    return this;
+  }
+
+  /** Returns properties if non-null; otherwise throws the appropriate exception. */
+  private PropertiesObject getPropertiesNonnull() throws CacheMetadataCorruptedException {
```
elharo: NonNull

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  public CacheMetadataLayerObjectTemplate setExistsOn(List<String> existsOn) {
+    this.existsOn = existsOn;
+    return this;
+  }
+
+  public CacheMetadataLayerObjectTemplate setSourceDirectories(List<String> sourceDirectories) {
+    checkTypeValidForProperties();
+    createPropertiesIfNull().sourceDirectories = sourceDirectories;
```
elharo: Huh? Please break this into two statements. It's confusing as is.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
```diff
@@ -0,0 +1,164 @@
...
+  public DescriptorDigest getDiffId() {
+    return reference.diffId;
+  }
+
+  public List<String> getExistsOn() {
+    return existsOn;
+  }
+
+  public List<String> getSourceDirectories() throws CacheMetadataCorruptedException {
+    return getPropertiesNonnull().sourceDirectories;
```
elharo: This exposes mutable internal state.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
```diff
@@ -0,0 +1,164 @@
...
+
+  public CacheMetadataLayerObjectTemplate setSourceDirectories(List<String> sourceDirectories) {
+    checkTypeValidForProperties();
+    createPropertiesIfNull().sourceDirectories = sourceDirectories;
+    return this;
+  }
+
+  public CacheMetadataLayerObjectTemplate setLastModifiedTime(long lastModifiedTime) {
+    checkTypeValidForProperties();
+    createPropertiesIfNull().lastModifiedTime = lastModifiedTime;
```
elharo: two statements please

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
```diff
@@ -0,0 +1,164 @@
...
+        && type != CachedLayerType.RESOURCES
+        && type != CachedLayerType.CLASSES) {
+      throw new IllegalStateException(
+          "Properties is not a valid field for non-application layer type");
+    }
+  }
+
+  /** Instantiates {@link #properties} if it's {@code null}. */
+  private PropertiesObject createPropertiesIfNull() {
+    if (null != properties) {
```
elharo: flip

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
coollog: I think this is fine here since this is just a template class and should be only temporary until translation to/from the real CacheMetadata is finished. Internal states of this template can be passed on to CacheMetadata without needing two copies otherwise.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
loosebazooka: Why is BASE a CachedLayerType and also stored in baseImageLayers?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataTemplate.java
```diff
@@ -0,0 +1,71 @@
...
+ *     },
+ *     ...
+ *     {
+ *       "type": "CLASSES",
+ *       "reference": {
+ *         "size": 223,
+ *         "digest": "sha256:8c662931926fa990b41da3c9f42663a537ccd498130030f9149173a0493832ad",
+ *         "diffId": "sha256:a3f3e99c29370df48e7377c8f9baa744a3958058a766793f821dadcb144a8372"
+ *       },
+ *       "existsOn": ["gcr.io/my-project/my-image"],
```
loosebazooka: existsOn sounds like it should be a runtime check?

loosebazooka:
Is it possible to separate out the renaming of JsonTemplateMapper into a different CL?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
coollog: This class is completely rewritten in #71 and will not have two separate fields for base layers and application layers.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataTemplate.java
coollog: This field keeps track of which image tags this layer has been pushed to, and will be used as an indicator as to whether or not a layer should be pushed. Usually, layers should not be deleted from a registry, and this should directly reflect the existence of the layer on the registry. If the layer has been deleted, however, this edge case can be detected when the final manifest is pushed (at which point it will error and we can clear the cache and rebuild).

Perhaps I should rename this `pushedTo`?

coollog:
Filed #73 for renaming to JsonTemplateMapper.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataTemplate.java
loosebazooka: How expensive is it to check if a layer exists remotely though? My only concern is that this optimization doesn't seem completely safe.

coollog: It's just extra HEAD requests. It should be safe since a registry won't accept a manifest with nonexistent layers.

loosebazooka: Oh, wait I misread that, are we going to handle it correctly for users? What do you mean by error?

coollog: When the registry receives a manifest with nonexistent layers, it will send back an error response.

When that happens, we can either:
- fail and tell the user to clear their cache and retry, or
- push all the layers and the manifest again

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
elharo: In that case, should we just merge this upstream and get rid of this PR?

PR #71, opened 2017-12-20
Adds JSON translator for CacheMetadata.

coollog:
Part of #52


coollog:
@elharo Merged #70 into here.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadataTranslator.java
```diff
@@ -0,0 +1,108 @@
...
+        switch (layerObjectTemplate.getType()) {
+          case BASE:
+            layerMetadata =
+                new LayerMetadata(layerObjectTemplate.getType(), layerObjectTemplate.getExistsOn());
+            break;
+
+          case DEPENDENCIES:
+          case RESOURCES:
+          case CLASSES:
+            layerMetadata =
```
elharo: Why can't it be this unconditionally?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/LayerMetadata.java
```diff
@@ -0,0 +1,71 @@
...
+
+  LayerMetadata(CachedLayerType type, List<String> existsOn) {
+    this.type = type;
+    this.existsOn = existsOn;
+  }
+
+  LayerMetadata(
+      CachedLayerType type,
+      List<String> existsOn,
+      @Nonnull List<String> sourceDirectories,
```
elharo: Why does this have to be nonnull if the field is Nullable?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
```diff
@@ -0,0 +1,167 @@
...
+    createPropertiesIfNull();
+    assert properties != null;
+    properties.sourceDirectories = sourceDirectories;
+    return this;
+  }
+
+  public CacheMetadataLayerObjectTemplate setLastModifiedTime(long lastModifiedTime) {
+    checkTypeValidForProperties();
+    createPropertiesIfNull();
+    assert properties != null;
```
elharo: Google style requires no asserts

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
```diff
@@ -0,0 +1,167 @@
...
+  private final ReferenceObject reference = new ReferenceObject();
+
+  /**
+   * A list of image tags where the layer should exist at (i.e. the layer was pushed to the
+   * repository for that image).
+   */
+  private List<String> existsOn = new ArrayList<>();
+
+  /** Additional properties for the layer. */
+  @Nullable private PropertiesObject properties;
```
elharo: You go to a lot of trouble to lazily initialize this. Could we just initialize this here?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadataTranslator.java
coollog: Base image layers do not have the fields for `#getSourceDirectories` and `#getLastModifiedTime` and the corresponding methods on `CacheMetadataLayerObjectTemplate` are intended to be only called for application layers (will throw an exception if those properties are not found).

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/LayerMetadata.java
coollog: This constructor is intended for application layers (as opposed to the other one being for base image layers). I'll add a comment explaining this.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/json/CacheMetadataLayerObjectTemplate.java
coollog: This field is intended to remain null for base image layers. The null-ness is used as a check for corrupted metadata where application layers lack a properties field.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadataTranslator.java
elharo: I suspected something like that, but in that case I'd expect these fields to return null, which could be passed in here. 

Otherwise, it feels like the class hierarchy is not quite right. Objects shouldn't have methods that don't make sense for some instances of the type. Any method on a class should be applicable to all instances of the type.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/Cache.java
```diff
@@ -0,0 +1,79 @@
...
+   * Initializes a cache with a directory. This also loads the cache metadata if it exists in the
+   * directory.
+   */
+  public static Cache init(File cacheDirectory)
+      throws NotDirectoryException, CacheMetadataCorruptedException {
+    if (!cacheDirectory.isDirectory()) {
+      throw new NotDirectoryException("The cache can only write to a directory");
+    }
+
+    CacheMetadata cacheMetadata = loadCacheMetadata(cacheDirectory.toPath());
```
elharo: would it make sense to make cacheDirectpry a Path instead of a File?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadataTranslator.java
```diff
@@ -0,0 +1,111 @@
...
+
+      // Converts each layer object in the template to a cache metadata layer.
+      for (CacheMetadataLayerObjectTemplate layerObjectTemplate : template.getLayers()) {
+        File layerContentFile =
+            CacheFiles.getLayerFile(cacheDirectory, layerObjectTemplate.getDigest());
+
+        // Gets the properties for a layer. Properties only exist for application layers.
+        CacheMetadataLayerPropertiesObjectTemplate propertiesObjectTemplate =
+            layerObjectTemplate.getProperties();
+        List<String> sourceDirectories = null;
```
elharo: empty list is sometimes safer than null

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/Cache.java
coollog: If `cacheDirectory` is a `Path`, then the check for `isDirectory` would need to first convert it to `File`. Is there a standard as to whether `Path` or `File` is preferred for a parameter?

elharo: No, you can use `Files.isDirectory(path)` instead. 

Generally we prefer Path for new code, but there's a lot of legacy code that will require a File.

coollog:
Need to remove "existsOn" field - not useful for now.

coollog:
@loosebazooka As discussed, the `existsOn` cache metadata field has been removed and separate caches are now used for the base image layers vs. application image layers.

PR #72, opened 2017-12-20
Adds JSON to image translator.

coollog:


crepecake/src/main/java/com/google/cloud/tools/crepecake/image/json/JsonToImageTranslator.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.cloud.tools.crepecake.image.LayerCountMismatchException;
+import com.google.cloud.tools.crepecake.image.LayerPropertyNotFoundException;
+import com.google.cloud.tools.crepecake.image.ReferenceLayer;
+import com.google.cloud.tools.crepecake.image.ReferenceNoDiffIdLayer;
+import java.util.ArrayList;
+import java.util.List;
+
+// TODO: Add translation of container configuration config.
+/** Translates {@link V21ManifestTemplate} and {@link V22ManifestTemplate} into {@link Image}. */
+public abstract class JsonToImageTranslator {
```
loosebazooka: private constructor?

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/Image.java
```diff
@@ -16,29 +16,33 @@
...
 /** Represents an image. */
 public class Image {
 
   /** The layers of the image, in the order in which they are applied. */
   private ImageLayers<Layer> layers = new ImageLayers<>();
 
-  /** Environment variables for running the image. Maps from variable name to its value. */
-  private final Map<String, String> environmentMap = new HashMap<>();
+  /** Environment variable definitions for running the image, in the format {@code NAME=VALUE}. */
+  private final List<String> environment = new ArrayList<>();
```
loosebazooka: why was this flipped? Are we defining our json different now?

coollog: This was changed because it is easier to just use the `NAME=VALUE` string rather than convert to-and-from a Map (i.e. existing `NAME=VALUE` strings do not need to be modified, and new `NAME=VALUE` strings that may be added can just be added as their string form).

elharo: This is the sort of thing I've found Guava is really good at if you know where to look. In particular see Splitter and Joiner. 

Also it suggests maybe we're missing an abstraction. I.e. it probably shouldn't be a List or a Map, but rather some sort of `Environment` class that hides these details from the client.

coollog: Filed https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/91 to do this abstraction.

coollog:
Yes, just awaiting final approval from @loosebazooka

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/json/JsonToImageTranslator.java
loosebazooka: just need this addressed and then we're good to go?

coollog: This has already been addressed via https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/pull/72/commits/b29fdc5214b5910ec96c30373c7f0c5829b2cd97

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/json/JsonToImageTranslator.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.cloud.tools.crepecake.image.Image;
+import com.google.cloud.tools.crepecake.image.Layer;
+import com.google.cloud.tools.crepecake.image.LayerCountMismatchException;
+import com.google.cloud.tools.crepecake.image.LayerPropertyNotFoundException;
+import com.google.cloud.tools.crepecake.image.ReferenceLayer;
+import com.google.cloud.tools.crepecake.image.ReferenceNoDiffIdLayer;
+import java.util.ArrayList;
+import java.util.List;
+
+// TODO: Add translation of container configuration config.
```
elharo: can the todo be removed now?

coollog: Yes, removed.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/json/JsonToImageTranslator.java
loosebazooka: ah okay :D

PR #79, opened 2017-12-27
Adds CacheWriter.

coollog:
Part of #52. Writes layers to the cache.


coollog:
Yes, this PR requires #71 to be merged first.

elharo:
Conflicts need to be resolved.

coollog:
This one should actually come after #84 now.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.util.zip.GZIPOutputStream;
+
+/** Writes {@link UnwrittenLayer}s to the cache. */
+public class CacheWriter {
+
+  private static final String TEMP_FILE_PREFIX = ".tmp.";
```
elharo: Why start with a dot? Do we want these to be hidden?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public CacheWriter(Cache cache) {
+    this.cache = cache;
+  }
+
+  public CachedLayer writeLayer(UnwrittenLayer layer) throws IOException {
+    // Writes to a temporary file first because the UnwrittenLayer needs to be written first to
+    // obtain its digest.
+    File tempLayerFile =
+        File.createTempFile(TEMP_FILE_PREFIX, TEMP_FILE_SUFFIX, cache.getCacheDirectory().toFile());
```
elharo: Files.createTempFile and then you can use paths

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
```diff
@@ -0,0 +1,75 @@
...
+    tempLayerFile.deleteOnExit();
+
+    // Writes the UnwrittenLayer layer BLOB to a file to convert into a CachedLayer.
+    try (CountingDigestOutputStream compressedDigestOutputStream =
+        new CountingDigestOutputStream(
+            new BufferedOutputStream(new FileOutputStream(tempLayerFile)))) {
+      // Writes the layer with GZIP compression. The original bytes are captured as the layer's
+      // diff ID and the bytes outputted from the GZIP compression are captured as the layer's
+      // content descriptor.
+      DescriptorDigest diffId;
```
elharo: can we move this into the try block?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
```diff
@@ -0,0 +1,75 @@
...
+          CacheFiles.getLayerFile(cache.getCacheDirectory(), compressedBlobDescriptor.getDigest());
+      if (!tempLayerFile.renameTo(layerFile)) {
+        throw new IOException(
+            "Could not rename layer "
+                + compressedBlobDescriptor.getDigest().getHash()
+                + " to "
+                + layerFile);
+      }
+
+      return new CachedLayer(layerFile, compressedBlobDescriptor, diffId);
```
elharo: I may be misreading this since I don't see a guarantee of diffId being initialized. Does this compile? Probably all of this should be in the try block.

crepecake/src/test/java/com/google/cloud/tools/crepecake/cache/CacheWriterTest.java
```diff
@@ -0,0 +1,105 @@
...
+  @Test
+  public void testWriteLayer() throws URISyntaxException, IOException {
+    File blobA = new File(Resources.getResource("blobA").toURI());
+    String expectedBlobAString =
+        new String(Files.readAllBytes(blobA.toPath()), StandardCharsets.UTF_8);
+
+    // Gets the expected content descriptor and diff ID.
+    CountingDigestOutputStream compressedDigestOutputStream =
+        new CountingDigestOutputStream(ByteStreams.nullOutputStream());
+    CountingDigestOutputStream uncompressedDigestOutputStream;
```
elharo: Your try blocks can be much larger and then you can move the declarations to the initialization.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
coollog: There is no requirement for them to be hidden. Is there a standard for naming of temporary files?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
coollog: The try block was just used to close the GZIPOutputStream. I removed it to just call close explicitly.

crepecake/src/test/java/com/google/cloud/tools/crepecake/cache/CacheWriterTest.java
coollog: Done.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
elharo: Not really. I wouldn't even bother with a prefix if possible.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
elharo: Using a try block to close is a very good idea. It's more reliable than closing explicitly since an IOException can be thrown before the close, an then the stream isn't closed. However you do need to put all the code that depends on the stream you declare in the try block into the try block.

coollog: Here, the `GzipCompressorStream` should be safe to not close since the `FileOutputStream` is already wrapped in a try block.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
```diff
@@ -0,0 +1,72 @@
...
+    tempLayerFile.deleteOnExit();
+
+    // Writes the UnwrittenLayer layer BLOB to a file to convert into a CachedLayer.
+    try (CountingDigestOutputStream compressedDigestOutputStream =
+        new CountingDigestOutputStream(
+            new BufferedOutputStream(new FileOutputStream(tempLayerFile)))) {
+      // Writes the layer with GZIP compression. The original bytes are captured as the layer's
+      // diff ID and the bytes outputted from the GZIP compression are captured as the layer's
+      // content descriptor.
+      GZIPOutputStream compressorStream = new GZIPOutputStream(compressedDigestOutputStream);
```
elharo: Could we just move this up into the chain? That would avoid the need to explicitly close.

coollog: Moving this up and not closing would mean that we would have to move code out of the try block again. The close must come before `compressedDigestOutputStream.toBlobDescriptor()` or else it doesn't capture the full compressed bytes.

elharo: OK, that makes sense. Might be worth adding a comment to that effect in case anyone in the future gets confused by this.

coollog: Done.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheWriter.java
```diff
@@ -0,0 +1,73 @@
...
+  private final Cache cache;
+
+  public CacheWriter(Cache cache) {
+    this.cache = cache;
+  }
+
+  public CachedLayer writeLayer(UnwrittenLayer layer) throws IOException {
+    // Writes to a temporary file first because the UnwrittenLayer needs to be written first to
+    // obtain its digest.
+    File tempLayerFile = Files.createTempFile(cache.getCacheDirectory(), null, null).toFile();
```
loosebazooka: If we delete on exit, does that not mess with what happens after renaming?

coollog: `deleteOnExit` called `DeleteOnExitHook.add(path)`, which adds the file path to a shutdown hook. When the hook is activated, the file is deleted with `(new File(filename)).delete();`, which doesn't do anything if the `filename` doesn't exist anymore, which is the case here after renaming.

PR #81, opened 2018-01-02
Adds CacheChecker.

coollog:
Part of #52. Checks if the cache is outdated.


crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Checks all cached layers built from the source files to see if the source files have been
+   * modified since the newest layer build.
+   *
+   * @param sourceFiles the source files to check
+   * @return true if no cached layer exists that are up-to-date with the source files; false
+   *     otherwise.
+   */
+  public boolean areSourceFilesModified(Set<File> sourceFiles)
```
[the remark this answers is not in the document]
loosebazooka: That's pretty expensive huh...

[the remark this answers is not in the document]
coollog: Yea we could check based on checksum, but that would involve reading the full file. Last modified time may not be completely reliable since other things can modify the last modified time, but that might be the tradeoff - we make the assumption that the files in our cache are only touched by us.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright 2017 Google Inc.
```
elharo: 2018

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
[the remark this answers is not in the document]
elharo: And how expensive is reading the full file in the environment where this will run. Can we estimate this in seconds, milliseconds, etc.?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    return true;
+  }
+
+  /**
+   * Checks the file has been modified since the {@code lastModifiedTime}. Recursively checks all
+   * subfiles if {@code file} is a directory.
+   */
+  private boolean isFileModifiedRecursive(File file, long lastModifiedTime) throws IOException {
```
elharo: isFileModified, don't need Recursive in name; or maybe even just isModified

Also, looks like this method can be static?

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,111 @@
...
+
+      // This layer is an up-to-date layer.
+      return false;
+    }
+
+    return true;
+  }
+
+  /**
+   * Checks the file has been modified since the {@code lastModifiedTime}. Recursively checks all
```
elharo: Checks the --> returns true if

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,111 @@
...
+   * subfiles if {@code file} is a directory.
+   */
+  private boolean isFileModifiedRecursive(File file, long lastModifiedTime) throws IOException {
+    if (file.lastModified() > lastModifiedTime) {
+      return true;
+    }
+
+    if (file.isDirectory()) {
+      File[] subFiles = file.listFiles();
+      if (null == subFiles) {
```
elharo: subFiles == null

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,111 @@
...
+        cache.getMetadata().filterLayers().bySourceFiles(sourceFiles).filter();
+
+    for (CachedLayerWithMetadata cachedLayer : cachedLayersWithSourceFiles) {
+      // Checks if the layer is outdated.
+      long lastModifiedTime = cachedLayer.getMetadata().getLastModifiedTime();
+      boolean hasOutdatedFile = false;
+      for (File file : sourceFiles) {
+        if (isFileModifiedRecursive(file, lastModifiedTime)) {
+          hasOutdatedFile = true;
+          break;
```
elharo: break and continue? maybe use a labeled continue here instead? Better yet, can't you simply immediately return true? That would make this a lot simpler.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
```diff
@@ -29,12 +33,87 @@
...
+
+    private boolean isSourceFilesFilterEnabled = false;
+
+    private LayerFilter(ImageLayers<CachedLayerWithMetadata> layers) {
+      this.layers = layers;
+    }
+
+    /** Filters to a certain layer type. */
+    LayerFilter byType(CachedLayerType type) {
+      this.type = type;
```
elharo: This is weird. I'd expect a method like this to return a new object, not modify the existing object. Maybe it's just the method name that's throwing me.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
```diff
@@ -29,12 +33,87 @@
...
+      isTypeFilterEnabled = true;
+      return this;
+    }
+
+    /** Filters to a certain set of source files. */
+    LayerFilter bySourceFiles(Set<File> sourceFiles) {
```
elharo: ditto

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.IOException;
+import java.util.Set;
+
+/** Checks if cached data is outdated. */
+public class CacheChecker {
+
+  private final Cache cache;
+
+  /**
+   * @return the last modified time for the file. Recursively finds the largest last modified time
```
elharo: largest --> latest or "most recent"

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,110 @@
...
+   *     for all subfiles if {@code file} is a directory.
+   */
+  private static long getLastModifiedTime(File file) throws IOException {
+    long lastModifiedTime = file.lastModified();
+
+    if (file.isDirectory()) {
+      File[] subFiles = file.listFiles();
+      if (subFiles == null) {
```
elharo: This should only happen if file is not a directory. Given that, you can simplify this as 

```
if (subFiles != null) {
  loop...
```

and not do the files.isDirectory check. 
  
Then again FileVisitor from NIO may be simpler for all this:

https://docs.oracle.com/javase/tutorial/essential/io/walk.html

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,110 @@
...
+  private final Cache cache;
+
+  /**
+   * @return the last modified time for the file. Recursively finds the largest last modified time
+   *     for all subfiles if {@code file} is a directory.
+   */
+  private static long getLastModifiedTime(File file) throws IOException {
+    long lastModifiedTime = file.lastModified();
+
+    if (file.isDirectory()) {
```
elharo: 1. What happens if this is a link? If so we could get stuck in an infinite loop. 
2. You probably want file.canRead() rather than isDirectory. There are unreadable directories due to permissions.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,110 @@
...
+      File[] subFiles = file.listFiles();
```
elharo: does listFiles return ..?
If so this gets stuck in an infinite loop.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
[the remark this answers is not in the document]
elharo: Thanks. The next question is will this be run in an environment where the difference between 0.5s and 14s is relevant. If it's a user facing job, yes. If a batch job probably not.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
elharo: What if someone passes null to these methods? That still sets isTypeFilterEnabled to true.

crepecake/src/test/java/com/google/cloud/tools/crepecake/cache/CacheCheckerTest.java
```diff
@@ -0,0 +1,167 @@
+/*
+ * Copyright 2017 Google Inc.
```
elharo: 2018

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -28,22 +30,21 @@
...
   private static long getLastModifiedTime(File file) throws IOException {
     long lastModifiedTime = file.lastModified();
 
-    if (file.isDirectory()) {
-      File[] subFiles = file.listFiles();
-      if (subFiles == null) {
-        throw new IOException("Failed to read directory: " + file.getAbsolutePath());
+    if (file.canRead()) {
+      OptionalLong maxLastModifiedTime =
+          Files.walk(file.toPath()).mapToLong(path -> path.toFile().lastModified()).max();
```
elharo: You don't need to convert to a file here. Just use Files.getLastModifiedTime(path)

In general converting to a java.io.File should not be necessary except when interfacing with older third party APIs (e.g. Eclipse APIs) that assume the use of java.io.File. All standard operations in the JDK should be able to be done with a. Path.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -28,22 +30,21 @@
   private final Cache cache;
 
   /**
-   * @return the last modified time for the file. Recursively finds the largest last modified time
-   *     for all subfiles if {@code file} is a directory.
+   * @return the last modified time for the file. Recursively finds the most recent last modified
+   *     time for all subfiles if {@code file} is a directory.
    */
   private static long getLastModifiedTime(File file) throws IOException {
```
elharo: Might want to return a FileTime rather than a long.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
```diff
@@ -29,12 +33,87 @@
...
+          if (isTypeFilterEnabled) {
+            if (type != layer.getMetadata().getType()) {
+              continue;
+            }
+          }
+
+          if (isSourceFilesFilterEnabled) {
+            List<String> cachedLayerSourceFilePaths = layer.getMetadata().getSourceFiles();
+            if (cachedLayerSourceFilePaths == null) {
+              if (sourceFiles != null) {
```
elharo: This is redundant with `if (isSourceFilesFilterEnabled)` above

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
```diff
@@ -29,12 +33,87 @@
...
+  /** Can be used to filter layers in the metadata. */
+  static class LayerFilter {
+
+    private final ImageLayers<CachedLayerWithMetadata> layers;
+
+    @Nullable private CachedLayerType type;
+    @Nullable private Set<File> sourceFiles;
+
+    /** True if the filters are used; false otherwise. */
+    private boolean isTypeFilterEnabled = false;
```
elharo: These fields now feel redundant. We could check if type and sourceFiles are null directly.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -28,22 +30,21 @@
...
-    if (file.isDirectory()) {
-      File[] subFiles = file.listFiles();
-      if (subFiles == null) {
-        throw new IOException("Failed to read directory: " + file.getAbsolutePath());
+    if (file.canRead()) {
+      OptionalLong maxLastModifiedTime =
+          Files.walk(file.toPath()).mapToLong(path -> path.toFile().lastModified()).max();
+      if (!maxLastModifiedTime.isPresent()) {
+        throw new IllegalStateException(
+            "Could not get last modified time for all files in directory '" + file + "'");
```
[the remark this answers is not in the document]
coollog: Yea, it will fail if just one file doesn't have a last modified time. Essentially, it means the logic of `!allFilesHaveLastModifiedTime`.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
coollog: Removed isTypeFilterEnabled.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheMetadata.java
coollog: Removed isTypeFilterEnabled.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -0,0 +1,131 @@
...
+   * @param sourceFiles the source files to check
+   * @return true if no cached layer exists that are up-to-date with the source files; false
+   *     otherwise.
+   */
+  public boolean areSourceFilesModified(Set<Path> sourceFiles)
+      throws IOException, CacheMetadataCorruptedException {
+    // Grabs all the layers that have matching source files.
+    ImageLayers<CachedLayerWithMetadata> cachedLayersWithSourceFiles =
+        cache.getMetadata().filterLayers().bySourceFiles(sourceFiles).filter();
+    if (cachedLayersWithSourceFiles.size() == 0) return true;
```
elharo: Google style is to never use single line if:

https://google.github.io/styleguide/javaguide.html#s4.1-braces

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/LayerMetadata.java
```diff
@@ -16,35 +16,41 @@
...
   /** The paths to the source files that the layer was constructed from. */
   private List<String> sourceFiles;
 
-  /** The last time the layer was constructed, or negative if unknown. */
-  private long lastModifiedTime;
+  /** The last time the layer was constructed. */
+  private final FileTime lastModifiedTime;
+
+  LayerMetadata(List<String> sourceFiles, FileTime lastModifiedTime) {
+    if (sourceFiles.size() == 0) {
```
elharo: `sourceFiles.isEmpty()` is clearer and sometimes faster than `.size() == 0`

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/LayerMetadata.java
```diff
@@ -16,35 +16,41 @@
...
   List<String> getSourceFiles() {
     return sourceFiles;
   }
 
-  public long getLastModifiedTime() {
+  public FileTime getLastModifiedTime() {
     return lastModifiedTime;
   }
+
+  @VisibleForTesting
```
elharo: I may be losing the thread, but is `VisibleForTesting` right here? This seems to be used in non-test code and not used in test code, but I could be mixing up different setSourceFiles methods.

coollog: `LayerMetadata` is intended to be an immutable class. Its `setSourceFiles` method is only used in `CacheCheckerTest` line 138. The other `setSourceFiles` method that is not just for testing is on `CacheMetadataLayerPropertiesObjectTemplate` used as the JSON template for `LayerMetadata`.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheChecker.java
```diff
@@ -107,7 +107,9 @@ public boolean areSourceFilesModified(Set<Path> sourceFiles)
     // Grabs all the layers that have matching source files.
     ImageLayers<CachedLayerWithMetadata> cachedLayersWithSourceFiles =
         cache.getMetadata().filterLayers().bySourceFiles(sourceFiles).filter();
-    if (cachedLayersWithSourceFiles.size() == 0) return true;
+    if (cachedLayersWithSourceFiles.size() == 0) {
```
elharo: isEmpty()

PR #82, opened 2018-01-02
Adds CacheReader.

coollog:
Part of #52. Reads layer content from the cache.


crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheReader.java
```diff
@@ -0,0 +1,56 @@
...
+   * @return
+   * @throws CacheMetadataCorruptedException
+   */
+  public File getLayerFile(CachedLayerType layerType, Set<Path> sourceFiles)
+      throws CacheMetadataCorruptedException {
+    switch (layerType) {
+      case DEPENDENCIES:
+      case RESOURCES:
+      case CLASSES:
+        CacheMetadata cacheMetadata = cache.getMetadata();
```
loosebazooka: Does the caller of getLayerFile know that their layerType is not supported? Or is the expectation that any layer can be supported.

I guess this comes down to a question of whether it's completely necessary to differentiate layers based on TYPE since they might all function in the same way?

coollog: Yeah, this is intended to be only used for application layers. Base image layers will not have a sourceFiles property (since they aren't built from source files) and thus do not fit into the purpose of this method. I changed the wording of the comment a bit to explain this better.

coollog: Actually, I'll just add an UnsupportedOperationException.

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheReader.java
```diff
@@ -0,0 +1,56 @@
+package com.google.cloud.tools.crepecake.cache;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheReader.java
```diff
@@ -0,0 +1,56 @@
...
+        ImageLayers<CachedLayerWithMetadata> cachedLayers =
+            cacheMetadata.filterLayers().byType(layerType).bySourceFiles(sourceFiles).filter();
+
+        // Finds the newest cached layer for the layer type.
+        FileTime newestLastModifiedTime = FileTime.from(Instant.MIN);
+        File newestLayerFile = null;
+        for (CachedLayerWithMetadata cachedLayer : cachedLayers) {
+          FileTime cachedLayerLastModifiedTime = cachedLayer.getMetadata().getLastModifiedTime();
+          if (cachedLayerLastModifiedTime.compareTo(newestLastModifiedTime) <= 0) {
+            continue;
```
elharo: Can you rewrite this without the continue by inverting the if? I find continue harder to read.

crepecake/src/test/java/com/google/cloud/tools/crepecake/cache/CacheReaderTest.java
```diff
@@ -0,0 +1,52 @@
+package com.google.cloud.tools.crepecake.cache;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheReader.java
coollog: Added copyright

crepecake/src/test/java/com/google/cloud/tools/crepecake/cache/CacheReaderTest.java
coollog: Added copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheReader.java
```diff
@@ -0,0 +1,72 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.cache;
+
+import com.google.cloud.tools.crepecake.image.ImageLayers;
+import sun.plugin.dom.exception.InvalidStateException;
```
elharo: wrong exception

crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheReader.java
loosebazooka: I think the question is really that is type truly necessary to pass into this method?

coollog: It is necessary to filter the cached layers in the metadata (line 35).

coollog:
@loosebazooka Yea, layers are independent of their content. Here, the cache is tied to the type of layer, however. We might want to just remove the 3 application layer types and just have the layers in `getLayerFile` filter by `sourceFiles` instead.

Issue #78, opened 2017-12-27
Change registry protocol default to HTTPS


coollog:
Remote registries should send over HTTPS. However, local registries may not have HTTPS enabled. 

For local registries, an 'allow_http' flag could allow the use of HTTP protocol.

This is in regards to #68, where the default protocol used for manifest pulling is HTTP.

PR #85, opened 2018-01-02
Adds BlobPuller.

coollog:
Step 2 of https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/59

Adds RegistryClient as interface to registry and changes ManifestPuller to be a provider. Adds BlobPuller as another provider.


crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/LocalRegistry.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  protected void after() {
+    try {
+      String stopRegistryCommand = "docker stop registry";
+      Runtime.getRuntime().exec(stopRegistryCommand).waitFor();
+
+      String removeRegistryContainerCommand = "docker rm -v registry";
+      Runtime.getRuntime().exec(removeRegistryContainerCommand).waitFor();
+
+    } catch (Exception ex) {
```
elharo: Be more specific about the exceptions thrown if possible. Effective Java recommends against catching a general Exception.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
```diff
@@ -0,0 +1,130 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.file.Path;
+import javax.annotation.Nullable;
+
+/** Interfaces with a registry. */
+public class RegistryClient {
+
+  // TODO: This should be configurable.
+  private static final String PROTOCOL = "http";
```
elharo: why not https?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
```diff
@@ -0,0 +1,130 @@
...
+      Response response = connection.get(builder.build());
+
+      return registryEndpointProvider.handleResponse(response);
+
+    } catch (HttpResponseException ex) {
+      switch (ex.getStatusCode()) {
+        case HttpURLConnection.HTTP_BAD_REQUEST:
+        case HttpURLConnection.HTTP_NOT_FOUND:
+          // The name or reference was invalid.
+          ErrorResponseTemplate errorResponse;
```
elharo: one line to combine declaration and initialization.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
```diff
@@ -0,0 +1,130 @@
...
+      if (authorization != null) {
+        builder.setAuthorization(authorization);
+      }
+      Response response = connection.get(builder.build());
+
+      return registryEndpointProvider.handleResponse(response);
+
+    } catch (HttpResponseException ex) {
+      switch (ex.getStatusCode()) {
+        case HttpURLConnection.HTTP_BAD_REQUEST:
```
elharo: depending on switch case fallthrough is extremely error prone.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryEndpointProvider.java
```diff
@@ -0,0 +1,23 @@
+package com.google.cloud.tools.crepecake.registry;
```
elharo: copyright

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
coollog: This will be taken care of in https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/78. Using HTTPS directly here doesn't work with local registries.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
coollog: The throw at the end of the this statement group prevents the fallthrough from happening.

elharo: It's not that it's used incorrectly now. It's that it will be used incorrectly in the future as the code evolves. For this reason most static analysis tools report this as an error. See Chapter 2 of van der Linden's Deep C Secrets for more context; but tldr don't use switch statements that depend on fallthrough.

coollog: Hmm, it doesn't seem that this code qualifies as fallthrough since there is no code between the labels. I'm not sure what would be an alternative solution here since there's no way to combine labels into an OR without putting a colon-case in-between?

elharo: You could use individual case statements or  better yet use if-else instead of switch.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryEndpointProvider.java
coollog: Done

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/ManifestPullerIntegrationTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google Inc.
+ * Copyright 2018 Google Inc.
```
loosebazooka: I think you need to keep the old copyright, here and everywhere else?

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/LocalRegistry.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package com.google.cloud.tools.crepecake.registry;
+
+import java.io.IOException;
+import org.junit.rules.ExternalResource;
+import org.junit.rules.TestRule;
+
+/** {@link TestRule} that runs a local registry. */
+class LocalRegistry extends ExternalResource {
```
loosebazooka: Can we customize the registry's storage location? so that if "STOP" fails it will still be in a unique temp location?

coollog: Made each registry be a unique container.

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/LocalRegistry.java
```diff
@@ -0,0 +1,71 @@
...
+  /** Starts the local registry. */
+  @Override
+  protected void before() throws Throwable {
+    String runRegistryCommand =
+        "docker run -d -p "
+            + port
+            + ":5000 --restart=always --name "
+            + containerName
+            + " registry:2";
+    Runtime.getRuntime().exec(runRegistryCommand).waitFor();
```
loosebazooka: I think we want to be using processbuilder everywhere instead of runtime exec.

loosebazooka:
@elharo, this is the standard gradle/maven structure. `src/test/java`...

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
loosebazooka: Agreed, this block might be more readable as if/else.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
```diff
@@ -0,0 +1,130 @@
...
+            registryErrorExceptionBuilder.addErrorEntry(errorEntry);
+          }
+
+          throw registryErrorExceptionBuilder.build();
+
+        case HttpURLConnection.HTTP_UNAUTHORIZED:
+        case HttpURLConnection.HTTP_FORBIDDEN:
+          throw new RegistryUnauthorizedException(ex);
+
+        case 307: // Temporary Redirect
```
loosebazooka: 307 has no const?

PR #88, opened 2018-01-03
 Adds ManifestPusher.

coollog:
Step 3 of #59


crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Connection.java
```diff
@@ -40,10 +42,15 @@
  */
 public class Connection implements Closeable {
 
-  private static final HttpRequestFactory HTTP_REQUEST_FACTORY =
-      new NetHttpTransport().createRequestFactory();
-
-  private HttpRequestFactory requestFactory = HTTP_REQUEST_FACTORY;
+  /**
+   * Do not use {@link NetHttpTransport}. It does process response errors properly. A new {@link
+   * ApacheHttpTransport} needs to be created for each connection because otherwise, HTTP connection
```
elharo: no comma after otherwise

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/json/ErrorEntryTemplate.java
```diff
@@ -19,6 +19,7 @@
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.google.cloud.tools.crepecake.json.JsonTemplate;
 
+// TODO: Should include detail field as well - need to have custom parser
```
elharo: issue or fix

coollog: Filed https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/113

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/LocalRegistry.java
```diff
@@ -71,6 +73,8 @@ protected void after() {
 
   /** Runs a command with naive tokenization by whitespace. */
   private void runCommand(String command) throws IOException, InterruptedException {
-    new ProcessBuilder(command.split(" ")).start().waitFor();
+    if (new ProcessBuilder(command.split(" ")).start().waitFor() != 0) {
```
loosebazooka: Splitting on " " is probably less safe than accepting a List<String> as input. I think runCommand should probably be accepting List<String>.

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Connection.java
```diff
@@ -40,10 +42,15 @@
...
-  private static final HttpRequestFactory HTTP_REQUEST_FACTORY =
-      new NetHttpTransport().createRequestFactory();
-
-  private HttpRequestFactory requestFactory = HTTP_REQUEST_FACTORY;
+  /**
+   * Do not use {@link NetHttpTransport}. It does not process response errors properly. A new {@link
+   * ApacheHttpTransport} needs to be created for each connection because otherwise HTTP connection
+   * persistence causes the connection to throw {@link NoHttpResponseException}.
+   *
+   * @see <a
```
loosebazooka: Is this a bug, or working as intended?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/ManifestPusher.java
```diff
@@ -0,0 +1,73 @@
...
+      RegistryEndpointProperties registryEndpointProperties,
+      V22ManifestTemplate manifestTemplate,
+      String imageTag) {
+    this.registryEndpointProperties = registryEndpointProperties;
+    this.manifestTemplate = manifestTemplate;
+    this.imageTag = imageTag;
+  }
+
+  @Override
+  public void buildRequest(Request.Builder builder) {
```
loosebazooka: This pattern seems a little strange. Is there maybe a better way to do this?

crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Connection.java
coollog: This is a working as intended for the library but a breaking feature for us. Essentially, NetHttpTransport adheres strictly to the RFC specs and breaks if it processes messages that don't fit the specs. Essentially, in this case, it throws an IOException on `401 Unauthorized` responses that don't have a `WWW-Authenticate` header, which is the case for certain Docker registry messages. I believe there are other Docker registry messages also don't follow the RFC specs exactly.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/ManifestPusher.java
coollog: Changed this to be `getContent` that just returns the content to send with the request.

PR #93, opened 2018-01-05
Adds BlobPusher.

coollog:
Step 4 of #59


crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/ManifestPullerIntegrationTest.java
```diff
@@ -31,19 +30,14 @@
...
   public void testPull_v21() throws IOException, RegistryException {
     RegistryClient registryClient = new RegistryClient(null, "localhost:5000", "busybox");
-    ManifestTemplate manifestTemplate = registryClient.pullManifest("latest");
+    V21ManifestTemplate manifestTemplate =
+        registryClient.pullManifest("latest", V21ManifestTemplate.class);
 
     Assert.assertEquals(1, manifestTemplate.getSchemaVersion());
-    V21ManifestTemplate v21ManifestTemplate = (V21ManifestTemplate) manifestTemplate;
-    Assert.assertTrue(0 < v21ManifestTemplate.getFsLayers().size());
+    Assert.assertTrue(0 < manifestTemplate.getFsLayers().size());
```
elharo: not possible to test the exact number?
If so, it's a little more conventional to swap these arguments of <

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/ManifestPuller.java
```diff
@@ -103,4 +94,38 @@ public String getActionDescription() {
...
+    switch (schemaVersion) {
+      case 1:
+        return manifestTemplateClass.cast(
+            JsonTemplateMapper.readJson(jsonString, V21ManifestTemplate.class));
+
+      case 2:
+        return manifestTemplateClass.cast(
+            JsonTemplateMapper.readJson(jsonString, V22ManifestTemplate.class));
+
+      case -1:
```
elharo: if else would likely be more obvious; especially since your cases are out of order which interferes with the compiler optimizations for switch

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryEndpointProvider.java
```diff
@@ -34,6 +35,9 @@
   @Nullable
   BlobHttpContent getContent();
 
+  /** @return a list of MIME types to pass as an {@code HTTP Accept} header */
```
elharo: nit: HTTP does not need to be in code

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/ImageTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google Inc.
+ * Copyright 2018 Google Inc.
```
elharo: don't change the copyright date if we've previously this file

crepecake/src/test/java/com/google/cloud/tools/crepecake/image/json/ImageToJsonTranslatorTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google Inc.
+ * Copyright 2018 Google Inc.
```
elharo: ditto

crepecake/src/integration-test/java/com/google/cloud/tools/crepecake/registry/ManifestPullerIntegrationTest.java
coollog: It might not be safe to assume a constant number for the number of layers in gcr.io/distroless/java.

crepecake/src/main/java/com/google/cloud/tools/crepecake/image/json/V21ManifestTemplate.java
```diff
@@ -55,6 +55,8 @@
  */
 public class V21ManifestTemplate extends ManifestTemplate {
 
+  public static final String MEDIA_TYPE = "application/vnd.docker.distribution.manifest.v1+json";
```
loosebazooka: do docker versioning names match up funny? or is 2.1 just compatible with all v1 images?

coollog: Yea, Docker's versioning only applies to Version 2 images. The v1 (or v2 in V22ManifestTemplate) refers to the Schema version, such that [Version 2 Schema 1 is v1](https://docs.docker.com/registry/spec/manifest-v2-1/) and Version 2 Schema 2 is v2.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobPuller.java
```diff
@@ -54,6 +56,11 @@ public BlobHttpContent getContent() {
     return null;
   }
 
+  @Override
+  public List<String> getAccept() {
+    return Collections.emptyList();
```
loosebazooka: Blob Puller accepts nothing? or everything?

coollog: Blobs are only sent as application/octet-streams, so the Accept header is optional/doesn't do anything here.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
```diff
@@ -74,7 +87,36 @@ public void pushManifest(V22ManifestTemplate manifestTemplate, String imageTag)
...
+  public boolean pushBlob(DescriptorDigest blobDigest, Blob blob)
+      throws IOException, RegistryException {
+    BlobPusher blobPusher = new BlobPusher(registryEndpointProperties, blobDigest, blob);
+
+    // POST /v2/<name>/blobs/uploads/?mount={blob.digest}
+    String locationString = callRegistryEndpoint(blobPusher.initializer());
+    if (locationString == null) {
+      return true;
+    }
+    URL location = new URL(locationString);
```
loosebazooka: Can you make location or locationString more descriptive? is it a registryBlobPushTargetLocation? or something?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
```diff
@@ -74,7 +87,36 @@ public void pushManifest(V22ManifestTemplate manifestTemplate, String imageTag)
...
+
+    // PATCH <Location> with BLOB
```
loosebazooka: can you elaborate on this comment a little? What is that new "location" you're  getting back from this call?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
```diff
@@ -74,7 +87,36 @@ public void pushManifest(V22ManifestTemplate manifestTemplate, String imageTag)
...
+    location = new URL(callRegistryEndpoint(blobPusher.writer(location)));
```
loosebazooka: Can we have a new URL var, not reuse location? Since it appears to be setting up the target for the next call at an unrelated location?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
coollog: Yea, it is the `Location` header to push the BLOB to. I added some comments to BlobPusher about the process. Would `patchLocation` (and `putLocation` below) be better?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
coollog: I added some comments to BlobPusher about the process. Essentially, the push is conducted in three stages:

1) POST - initializes the upload - gets a Location back to send the BLOB content to
2) PATCH - write the BLOB to the upload Location, gets a Location back to commit the BLOB digest
3) PUT - commit the BLOB to a specific digest id

PR #95, opened 2018-01-05
Adds AuthenticationMethodRetriever.

### LINKED ISSUE #78: Change registry protocol default to HTTPS

### LINKED ISSUE #94: Add registry authenticator for other registries.
In `RegistryAuthenticators`, there is currently `RegistryAuthenticator#forDockerHub`.

Add `RegistryAuthenticator#forOther` that:

1) Fetches GET /v2/
2) Extracts the WWW-Authenticate header (eg. `WWW-Authenticate: Bearer realm="https://gcr.io/v2/token",service="gcr.io"`)
3) Parses the header and authenticates accordingly.

coollog:
Step 6 of #59
  
Fixes #78

Fixes #94

### DIFF
--- a/crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/AuthenticationMethodRetriever.java
+++ b/crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/AuthenticationMethodRetriever.java
@@ -17,6 +17,8 @@
 package com.google.cloud.tools.crepecake.registry;
 
 import com.google.api.client.http.HttpMethods;
+import com.google.api.client.http.HttpResponseException;
+import com.google.api.client.http.HttpStatusCodes;
 import com.google.cloud.tools.crepecake.http.BlobHttpContent;
 import com.google.cloud.tools.crepecake.http.Response;
 import java.net.MalformedURLException;
@@ -26,7 +28,7 @@ import java.util.List;
 import javax.annotation.Nullable;
 
 /** Retrieves the {@code WWW-Authenticate} header from the registry API. */
-class AuthenticationMethodRetriever implements RegistryEndpointProvider<Void> {
+class AuthenticationMethodRetriever implements RegistryEndpointProvider<RegistryAuthenticator> {
 
   private final RegistryEndpointProperties registryEndpointProperties;
 
@@ -42,8 +44,10 @@ class AuthenticationMethodRetriever implements RegistryEndpointProvider<Void> {
   }
 
   @Override
-  public Void handleResponse(Response response) {
-    return null;
+  public RegistryAuthenticator handleResponse(Response response) throws RegistryErrorException {
+    throw new RegistryErrorExceptionBuilder(getActionDescription())
+        .addReason("Did not receive '401 Unauthorized' response")
+        .build();
   }
 
   @Override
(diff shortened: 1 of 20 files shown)


crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/RegistryClient.java
```diff
@@ -50,6 +51,42 @@ public RegistryClient(@Nullable Authorization authorization, String serverUrl, S
...
+  /** Gets the {@link RegistryAuthenticator} to authenticate pulls from the registry. */
+  public RegistryAuthenticator getRegistryAuthenticator()
+      throws IOException, RegistryException, RegistryAuthenticationFailedException {
+    // Gets the WWW-Authenticate header (eg. 'WWW-Authenticate: Bearer realm="https://gcr.io/v2/token",service="gcr.io"')
+    AuthenticationMethodRetriever authenticationMethodRetriever =
+        new AuthenticationMethodRetriever(registryEndpointProperties);
+
+    try {
+      callRegistryEndpoint(authenticationMethodRetriever);
+      throw new RegistryErrorExceptionBuilder(authenticationMethodRetriever.getActionDescription())
```
loosebazooka: It seems weird for us to have code that is expecting an exception?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/AuthenticationMethodRetriever.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @Override
+  public List<String> getAccept() {
+    return Collections.emptyList();
+  }
+
+  @Override
+  public RegistryAuthenticator handleResponse(Response response) throws RegistryErrorException {
+    throw new RegistryErrorExceptionBuilder(getActionDescription())
```
loosebazooka: Maybe I'm missing something, but this seems to be always throwing this exception?

coollog: Yea, the RegistryAuthenticator expects a `401 Unauthorized` HTTP error, so if it receives a 200 status response (handled by `handleResponse`), it throws an exception.

loosebazooka: Oh I see, this just picks up the authentication method... is this really the only way to do it? Isn't there an endpoint to hit to ask?

coollog: Yeah, this is doing the authentication according to https://docs.docker.com/registry/spec/auth/token/, where:

1) Get the authentication method from the WWW-Authenticate header
2) Authenticate with the provided method (done by `RegistryAuthenticator`)

loosebazooka: Damn that's so silly. okay :smile:

coollog: Here, we are doing it once before all pull calls. But, we might want to have this be done every time we receive a 401 Unauthorized, so I filed https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/96

loosebazooka: Not sure if you linked to that somewhere, but it'd be nice to include in the code so someone looking at this in the future will understand.

loosebazooka: just kdiding I found it in the code...

coollog: Yeah, it is linked in `RegistryAuthenticator`. `AuthenticationMethodRetriever` just does standard `WWW-Authenticate` header retrieval.

Issue #60, opened 2017-12-13
Add build steps that pulls, builds, and pushes from the components


coollog:
- [x] authenticate pull #105
- [x] pull base image manifest #105
- [x] pull base image layers #105
- [x] get push credentials #107 
- [x] push base image layers #107 
- [x] build application layers #109 
- [x] push application layers #112 
- [x] push new image manifest #112 
- [x] write-back new cache metadata #112 

Make sure to share no writable memory between these steps so that they can be easily parallelized.
  #

PR #97, opened 2018-01-05
Sets up builder configuration and step interface.

coollog:
Prepares for #60 .

`BuildConfiguration` is used to inject configuration for the image building process.

`Step` is an interface that will be implemented by the various steps in the image building process.


crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/BuildConfiguration.java
```diff
@@ -0,0 +1,135 @@
...
+      return this;
+    }
+
+    /** Sets the server URL of the registry to push the built image to. */
+    public Builder setTargetServerUrl(String targetServerUrl) {
+      this.targetServerUrl = targetServerUrl;
+      return this;
+    }
+
+    /** Sets the image name/repository the built image (also known as the registry namespace). */
```
elharo: of the built image?

crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/BuildConfiguration.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /** Sets the credential helper name used by {@link DockerCredentialRetriever}. */
+    public Builder setCredentialHelperName(String credentialHelperName) {
+      this.credentialHelperName = credentialHelperName;
+      return this;
+    }
+
+    public BuildConfiguration build() {
+      return new BuildConfiguration(
```
elharo: These are all nullable?

crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/Step.java
```diff
@@ -0,0 +1,24 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.crepecake.builder;
+
+/** A step in the builder process. Implementations must be thread-safe. */
+@FunctionalInterface
+interface Step<T, R> {
+
+  R run(T input) throws Exception;
```
elharo: probably a bad idea for functional methods to throw exceptions; and we really shouldn't throw an arbitrary exception

crepecake/src/test/java/com/google/cloud/tools/crepecake/builder/BuildConfigurationTest.java
```diff
@@ -0,0 +1,44 @@
...
+    BuildConfiguration buildConfiguration =
+        BuildConfiguration.builder()
+            .setBaseImageServerUrl(expectedBaseImageServerUrl)
+            .setBaseImageName(expectedBaseImageName)
+            .setTargetServerUrl(expectedTargetServerUrl)
+            .setTargetImageName(expectedTargetImageName)
+            .setTargetTag(expectedTargetTag)
+            .setCredentialHelperName(expectedCredentialHelperName)
+            .build();
+    Assert.assertEquals(expectedBaseImageName, buildConfiguration.getBaseImageName());
```
elharo: should test the other fields too

crepecake/src/test/java/com/google/cloud/tools/crepecake/builder/BuildConfigurationTest.java
```diff
@@ -0,0 +1,44 @@
...
+package com.google.cloud.tools.crepecake.builder;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class BuildConfigurationTest {
+
+  @Test
+  public void testBuilder() {
+    String expectedBaseImageServerUrl = "someserver";
```
elharo: should test when each of these is null

crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/BuildConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+      values.put(Fields.TARGET_TAG, targetTag);
+      return this;
+    }
+
+    public Builder setCredentialHelperName(String credentialHelperName) {
+      values.put(Fields.CREDENTIAL_HELPER_NAME, credentialHelperName);
+      return this;
+    }
+
+    public BuildConfiguration build() throws BuildConfigurationMissingValueException {
```
elharo: Per Effective Java 2e, Item 60, you should use an IllegalStateException for this case.

crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/BuildConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+      BuildConfigurationMissingValueException.Builder
+          buildConfigurationMissingValueExceptionBuilder =
+              BuildConfigurationMissingValueException.builder();
+      for (Fields field : Fields.values()) {
+        if (!values.containsKey(field)) {
```
elharo: You could put this info in the exception message of an IllegalStateException.

crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/BuildConfiguration.java
```diff
@@ -0,0 +1,177 @@
...
+            } else {
+              stringBuilder.append(", ");
+            }
+            stringBuilder.append(descriptions.get(descriptionsIndex));
+          }
+          throw newIllegalStateException(stringBuilder.toString());
+      }
+    }
+
+    private IllegalStateException newIllegalStateException(String requiredDescriptions) {
```
elharo: inline this. It's a one liner.

crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/BuildConfiguration.java
```diff
@@ -0,0 +1,177 @@
...
+        case 0:
+          values = Collections.unmodifiableMap(values);
+          return new BuildConfiguration(values);
+
+        case 1:
+          throw new IllegalStateException(descriptions.get(0) + MISSING_FIELD_MESSAGE_SUFFIX);
+
+        case 2:
+          throw new IllegalStateException(
+              descriptions.get(0) + " and " + descriptions.get(1) + MISSING_FIELD_MESSAGE_SUFFIX);
```
elharo: I feel like there should be a utility method for this formatting pattern somewhere but I can't find it right now.

coollog: Added a TODO to find one.

crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/BuildConfiguration.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+
+/** Immutable configuration options for the builder process. */
+public class BuildConfiguration {
+
+  /** Enumerates the fields in the configuration. */
+  @VisibleForTesting
+  enum Fields {
```
loosebazooka: This looks like a weird way to use Enums. Since the values of "Fields" aren't related in any way. I would normally expect an enum to represent a range of values for a single type? Looks like we're just using it as a way to represent contants?

Perhaps just make them members of BuildConfiguration directly?

baseImageServerUrl... etc ?

coollog: This was mostly to be able to avoid a large set of `if` statements in Builder#build, since the `EnumMap` allows us to loop over the enum types. Perhaps a normal class would be better? I think later on I will add a required flag to the configurations as well.

PR #104, opened 2018-01-08
Throws IllegalStateException when trying to rewrite InputStreamBlob.

### LINKED ISSUE #103: Throw IllegalStateException when trying to rewrite InputStreamBlob.
`InputStreamBlobs` should only be allowed to be written out once (it closes the `InputStream`).

coollog:
Fixes #103

### DIFF
--- a/crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobWriter.java
+++ b/crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobWriter.java
@@ -19,6 +19,10 @@ package com.google.cloud.tools.crepecake.blob;
 import java.io.IOException;
 import java.io.OutputStream;
 
+/**
+ * Function that writes the contents of a BLOB. This functions should write the same BLOB regardless
+ * of the number of times it is called.
+ */
 @FunctionalInterface
 public interface BlobWriter {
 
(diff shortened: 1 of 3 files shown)


crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobWriter.java
```diff
@@ -19,6 +19,10 @@
 import java.io.IOException;
 import java.io.OutputStream;
 
+/**
+ * Function that writes the contents of a BLOB. This functions should write the same BLOB regardless
```
elharo: This function

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobWriter.java
```diff
@@ -19,6 +19,10 @@
...
+ * of the number of times it is called.
```
elharo: Isn't the point of this that it can only be called once?

coollog: `FileBlob`, `StringBlob`, and `WriterBlob` should all be able to be written out more than once. However, `InputStreamBlob` is tied to a read-once InputStream, so it can only be written out once.

crepecake/src/main/java/com/google/cloud/tools/crepecake/blob/BlobWriter.java
```diff
@@ -19,6 +19,10 @@
 import java.io.IOException;
 import java.io.OutputStream;
 
+/**
+ * This function that writes the contents of a BLOB. This functions should write the same BLOB
```
elharo: functions --> function

PR #105, opened 2018-01-08
Adds build steps for pulling the base image.

coollog:
Steps 1, 2, and 3 of #60 

These build steps will eventually go into the full build process. In this PR, the build steps that are added are:

1. Authenticate pull base image.
2. Pull the base image manifest.
3. Pull the base image layers.


crepecake/src/main/java/com/google/cloud/tools/crepecake/cache/CacheFiles.java
```diff
@@ -30,7 +30,8 @@ static Path getMetadataFile(Path cacheDirectory) {
   }
 
   /**
-   * Gets the file for a layer digest. The file is {@code <cache directory>/<layer hash>.tar.gz}.
+   * Gets the path to the file for a layer digest. The file is {@code <cache directory>/<layer
```
elharo: try not to use angle brackets here since it will be interpreted as markup.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
```diff
@@ -0,0 +1,129 @@
...
+  public BlobDescriptor handleHttpResponseException(HttpResponseException httpResponseException)
+      throws RegistryErrorException, HttpResponseException {
+    if (httpResponseException.getStatusCode() != HttpStatusCodes.STATUS_CODE_NOT_FOUND) {
+      throw httpResponseException;
+    }
+
+    // Finds a BLOB_UNKNOWN error response code.
+    String errorContent = httpResponseException.getContent();
+    if (errorContent == null) {
+      // TODO: The Google HTTP client gives null content for HEAD requests. Make the content never be null, even for HEAD requests.
```
elharo: issue? or fix now

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
```diff
@@ -0,0 +1,129 @@
...
+      return null;
+    } else {
+      try {
+        ErrorResponseTemplate errorResponse =
+            JsonTemplateMapper.readJson(errorContent, ErrorResponseTemplate.class);
+        List<ErrorEntryTemplate> errors = errorResponse.getErrors();
+        if (errors.size() == 1) {
```
elharo: can errors.size() be greater than 1?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobPuller.java
```diff
@@ -17,55 +17,43 @@
...
-  @Nullable
-  @Override
-  public BlobHttpContent getContent() {
-    return null;
-  }
-
-  @Override
-  public List<String> getAccept() {
-    return Collections.emptyList();
+    this.destOutputStream = destOutputStream;
```
elharo: perhaps rename "destination"

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
coollog: Filed https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/116

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
coollog: Yes, we only return null if there is one error and it is a BLOB_UNKNOWN error.

coollog:
@loosebazooka Good to go?

crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/Step.java
```diff
@@ -16,9 +16,19 @@
...
+import com.google.cloud.tools.crepecake.registry.RegistryAuthenticationFailedException;
+import com.google.cloud.tools.crepecake.registry.RegistryException;
+import java.io.IOException;
+
 /** A step in the builder process. Implementations must be thread-safe. */
 interface Step<T, R> {
 
   // TODO: Add more exceptions as needed.
-  R run(T input);
+  R run(T input)
```
loosebazooka: Since we doesn't seem like we're using Step in a List<Step> or anything, perhaps this interface doesn't make sense, especially since Run seems to inflating the exception thrown list based on each individual Step's impl.

coollog: Removing this in favor of implementing `Callable` to fit with `ExecutorService#submit`.

PR #107, opened 2018-01-09
Adds build steps for pushing the base image layers.

coollog:
Steps 4 and 5 of #60 .

These build steps will eventually go into the full build process. In this PR, the build steps that are added are:

1. Get credentials to push to the target registry.
2. Push the base image layers to the target registry.

### DIFF
--- a/crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/PushBaseImageLayersStep.java
+++ b/crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/PushBaseImageLayersStep.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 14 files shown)


crepecake/src/main/java/com/google/cloud/tools/crepecake/builder/PushBaseImageLayersStep.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  public Void call() throws IOException, RegistryException {
+    RegistryClient registryClient =
+        new RegistryClient(
+            pushAuthorization,
+            buildConfiguration.getTargetServerUrl(),
+            buildConfiguration.getTargetImageName());
+
+    // TODO: Pushing any BLOB should be in a separate build step.
```
coollog: Filed https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/117

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/DockerCredentialRetriever.java
```diff
@@ -82,13 +83,27 @@ public Authorization retrieve()
         // Throws an exception if the credential store does not have credentials for serverUrl.
         if (output.contains("credentials not found in native keychain")) {
```
loosebazooka: can we determine this by exit code?

coollog: We actually need to find an alternative to using the Docker credential helper method here, as implementations of the credential helper differ. Filed https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/118

PR #108, opened 2018-01-09
Adds BlobChecker.

coollog:
Step 5 of #59 .

The `BlobChecker` is an interface to the registry endpoint for checking if a BLOB exists on a registry or not. This endpoint returns the size of the BLOB in the `Content-Length` header if the BLOB exists, or gives a `404 Not Found` error (with error details in the response body) if the BLOB doesn't exist.

See also https://docs.docker.com/registry/spec/api/#existing-layers

### DIFF
--- a/crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
+++ b/crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
@@ -42,18 +42,19 @@ public class Response {
     return httpResponse.getHeaders().getHeaderStringValues(headerName);
   }
 
-  /**
-   * @return the first {@code Content-Length} header, or {@code -1} if not found
-   * @throws NumberFormatException if the {@code Content-Length} header could not be parsed as a
-   *     number
-   */
+  /** @return the first {@code Content-Length} header, or {@code -1} if not found */
   public long getContentLength() throws NumberFormatException {
     String contentLengthHeader =
         httpResponse.getHeaders().getFirstHeaderStringValue(HttpHeaders.CONTENT_LENGTH);
     if (contentLengthHeader == null) {
       return -1;
     }
-    return Long.parseLong(contentLengthHeader);
+    try {
+      return Long.parseLong(contentLengthHeader);
+
+    } catch (NumberFormatException ex) {
+      return -1;
+    }
   }
 
   /** Gets the HTTP response body as a {@link Blob}. */
--- a/crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
+++ b/crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
@@ -42,8 +42,12 @@ public class Response {
     return httpResponse.getHeaders().getHeaderStringValues(headerName);
   }
 
-  /** Gets the first {@code Content-Length} header, or {@code -1} if not found. */
-  public long getContentLength() {
+  /**
+   * @return the first {@code Content-Length} header, or {@code -1} if not found
+   * @throws NumberFormatException if the {@code Content-Length} header could not be parsed as a
+   *     number
+   */
+  public long getContentLength() throws NumberFormatException {
     String contentLengthHeader =
         httpResponse.getHeaders().getFirstHeaderStringValue(HttpHeaders.CONTENT_LENGTH);
     if (contentLengthHeader == null) {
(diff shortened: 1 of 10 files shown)


crepecake/src/main/java/com/google/cloud/tools/crepecake/http/Response.java
```diff
@@ -41,6 +42,16 @@ public int getStatusCode() {
...
   }
 
+  /** Gets the first {@code Content-Length} header, or {@code -1} if not found. */
+  public long getContentLength() {
+    String contentLengthHeader =
+        httpResponse.getHeaders().getFirstHeaderStringValue(HttpHeaders.CONTENT_LENGTH);
+    if (contentLengthHeader == null) {
+      return -1;
+    }
+    return Long.parseLong(contentLengthHeader);
```
loosebazooka: I wonder if we should catch the number format exception thrown by parseLong? In case the response isn't according to spec?

loosebazooka: or is that simply just not a possibility?

coollog: Added a check just in case.

elharo: It's totally a possibility, and we should probably return -1 in this case. Don't bubble the exception here.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
```diff
@@ -0,0 +1,136 @@
...
+    this.registryEndpointProperties = registryEndpointProperties;
+    this.blobDigest = blobDigest;
+  }
+
+  /** @return the BLOB's content descriptor */
+  @Override
+  public BlobDescriptor handleResponse(Response response) throws RegistryErrorException {
+    try {
+      long contentLength = response.getContentLength();
+      if (contentLength < 0) {
```
elharo: Content-Length is optional in HTTP, and not all servers send it all the time. Is it required and expected of the docker registry?

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
```diff
@@ -0,0 +1,136 @@
...
+  public BlobDescriptor handleHttpResponseException(HttpResponseException httpResponseException)
+      throws RegistryErrorException, HttpResponseException {
+    if (httpResponseException.getStatusCode() != HttpStatusCodes.STATUS_CODE_NOT_FOUND) {
+      throw httpResponseException;
+    }
+
+    // Finds a BLOB_UNKNOWN error response code.
+    String errorContent = httpResponseException.getContent();
+    if (errorContent == null) {
+      // TODO: The Google HTTP client gives null content for HEAD requests. Make the content never be null, even for HEAD requests.
```
elharo: file issue or fix here

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
```diff
@@ -0,0 +1,136 @@
...
+
+    } catch (NumberFormatException ex) {
+      throw new RegistryErrorExceptionBuilder(getActionDescription(), ex)
+          .addReason("Content-Length header was not a number")
+          .build();
+    }
+  }
+
+  @Override
+  public BlobDescriptor handleHttpResponseException(HttpResponseException httpResponseException)
```
elharo: This might be clearer if it were rewritten simply as `isBlobUnknownError` and the details of what to throw or return were left to a higher layer up the stack.

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
coollog: Yes, this is part of the Docker registry response. It looks like:

```
200 OK
Content-Length: <length>
Docker-Content-Digest: <digest>
Content-Type: application/octet-stream
```

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
coollog: Filed https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/116

crepecake/src/main/java/com/google/cloud/tools/crepecake/registry/BlobChecker.java
coollog: Do you mean to have isBlobUnknownError handle `404 Not Found` status codes so that there is no re-throwing of the `HttpResponseException` parameter?

PR #122, opened 2018-01-29
Removes crepecake.

coollog:
Moved to https://github.com/google/jib as Jib.


elharo:
You should probably make the repo private instead, and perhaps delete it.

coollog:
@elharo This repo is still used for the `minikube-build-tools-for-java`, such as the `minikube-gradle-plugin`.

PR #132, opened 2018-02-26
Fixes prepare_release script increment regex match.

coollog:
This does correct patch version incrementation with suffixes such as `-alpha.1`.


coollog:
Actually, this should not do this. When we release a version with a suffix, we should have the version be set back to the previous SNAPSHOT version.

For example, release `1.0.0-alpha.1` should set the pom.xml version back to `1.0.0-SNAPSHOT`.

PR #158, opened 2019-06-11
Populate Sonatype credentials

chanseokoh:


minikube-maven-plugin/kokoro/release_publish.sh
```diff
@@ -3,6 +3,8 @@
 # Fail on any error.
 set -e
 
+SONATYPE_USERNAME=$(cat "${KOKORO_KEYSTORE_DIR}/72743_sonatype_username")
```
loosebazooka: I'm not sure I understand the implications of this.

chanseokoh: We were injecting these env variables after externally populating the content of `72743_sonatype_username`, but this external injection is unnecessary. The file `72743_sonatype_username` is already present and the script can just query the content like this.

chanseokoh: And then we can remove the setup of the external unnecessary env var injection.

loosebazooka: I think @suztomo is reworking this whole release process, no?

chanseokoh: Yeah, I'm in touch with @suztomo about this.

